PR #1, opened 2022-03-31
SC-2040: create schema for postgres db

akashkj:
Changes to create schema in the DB. This is a sub-task under automatic roles creation US.

TODO - Adding unit test for postgres specific code, I'm looking at some mock utility lib that can be used for postgres testing.

### DIFF
--- a/hq_superset/model/db_info.py
+++ b/hq_superset/model/db_info.py
@@ -1,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class DbInfo(object):
+    database_name: str
+    db_user: str
+    db_password: str
+    db_host: str = "localhost"
+    db_port: str = "5432"
--- a/hq_superset/service/db/db_manager.py
+++ b/hq_superset/service/db/db_manager.py
@@ -1,0 +1,22 @@
+from abc import abstractmethod
+from typing import Any
+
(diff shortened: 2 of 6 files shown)


hq_superset/model/db_info.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class DbInfo(object):
```
sravfeyn, line 5: Can we not use the URL style connection string to avoid this class?

proteusvacuum (changes requested):

kaapstorm (commented):
It looks like we are trying to reinvent a very well-established wheel. Is there a reason we aren't using SQLAlchemy + Alembic [as Superset already does](https://github.com/apache/superset/blob/2631558ac4ba596b24876a9b0702bbf3a6bfbb38/superset/migrations/alembic.ini#L20) for managing our database?

sravfeyn:
Actually, this could be much simpler. I have created a util here to create a schema https://github.com/dimagi/hq_superset/pull/6/commits/c35339e28dbe2130da34fda24e248b8e9094bb16

akashkj:
That looks much simpler indeed @sravfeyn 
Will close this one

PR #2, opened 2022-04-01
Force a domain selection for certain URLs and maintain the domain con…

sravfeyn:
@proteusvacuum @akashkj @kaapstorm 

This implements the domain based multitenancy via cookie.

- It defines a `DOMAIN_REQUIRED_VIEWS` list which should require a domain selection.
- A domain list/selection view is created to let user pick a domain.
- Each request is intercepted before the processing, and if the matching view is one of `DOMAIN_REQUIRED_VIEWS` and the request cookie doesn't have `hq_domain`, the client is redirected to a domain selection view.
- When the user selects a domain, the `hq_domain` cookie value is set to that domain and all the subsequent requests will have that domain

To-dos
- A domain selection dropdown needs to be created.
- Fill `DOMAIN_REQUIRED_VIEWS` with exhaustive list of views
- Use correct list of user's domains once these tasks are implemented. https://dimagi-dev.atlassian.net/browse/SC-1998 and https://dimagi-dev.atlassian.net/browse/SC-2019

### DIFF
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -8,4 +8,11 @@ def patch_superset_config(config):
 def flask_app_mutator(app):
     # Import the views (which assumes the app is initialized) here
     # return
-    pass
+    from . import hq_domain
+    from . import views
+    from superset.extensions import appbuilder
+    appbuilder.add_view(views.SelectDomainView, 'Select Domain')
+    app.before_request_funcs.setdefault(None, []).append(
+        hq_domain.before_request_hook
+    )
+    app.strict_slashes = False
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -7,9 +7,7 @@ from flask_login import current_user
 
 def before_request_hook():
     override_jinja2_template_loader()
-    # Check if a hq_domain cookie is set
-    # If not redirect to domain_list view
-    return redirect_to_select_domain_view()
+    return ensure_domain_selected()
 
 
 def override_jinja2_template_loader():
@@ -37,7 +35,10 @@ DOMAIN_EXCLUDED_VIEWS = [
     "appbuilder.static",
 ]
 
-def redirect_to_select_domain_view():
+def ensure_domain_selected():
+    # Check if a hq_domain cookie is set
+    #   Ensure necessary roles, permissions and DB schemas are created for the domain
+    import superset
     if request.url_rule.endpoint in DOMAIN_EXCLUDED_VIEWS:
         return
     hq_domain = request.cookies.get('hq_domain')
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -2,22 +2,18 @@ import os
 
 import jinja2
 
-from flask import flash, g, redirect, request, url_for
-
-DOMAIN_REQUIRED_VIEWS = [
-    'TableModelView.list'
-]
-
+from flask import flash, g, redirect, request, url_for, session
+from flask_login import current_user
 
 def before_request_hook():
-    # Check URL is one of DOMAIN_REQUIRED_VIEWS
+    override_jinja2_template_loader()
     # Check if a hq_domain cookie is set
     # If not redirect to domain_list view
-    _override_jinja2_template_loader()
-    _validate_domain()
+    return redirect_to_select_domain_view()
 
 
-def _override_jinja2_template_loader():
+def override_jinja2_template_loader():
+    # Allow loading templates from the templates directory in this project as well
     from superset import app
 
     template_path = os.sep.join((
@@ -31,18 +27,29 @@ def _override_jinja2_template_loader():
     app.jinja_loader = my_loader
 
 
-def _validate_domain():
-    if not request.url_rule.endpoint in DOMAIN_REQUIRED_VIEWS:
+DOMAIN_EXCLUDED_VIEWS = [
+    "AuthOAuthView.login",
+    "AuthOAuthView.logout",
+    "AuthOAuthView.oauth_authorized",
+    "AuthDBView.logout",
+    "AuthDBView.login",
+    "SelectDomainView.list",
+    "appbuilder.static",
+]
+
+def redirect_to_select_domain_view():
+    if request.url_rule.endpoint in DOMAIN_EXCLUDED_VIEWS:
         return
+    hq_domain = request.cookies.get('hq_domain')
+    valid_domains = user_domains(current_user)
+    # Todo; Handle no superset enabled domains case
+    if hq_domain in valid_domains:
+        g.hq_domain = hq_domain
+    elif len(valid_domains) == 1:
+        g.hq_domain = valid_domains[0]
     else:
-        hq_domain = request.cookies.get('hq_domain')
-        if hq_domain and is_valid_user_domain(hq_domain):
-            import pdb; pdb.set_trace()
-            g.hq_domain = hq_domain
-            # add_domain_links(hq_domain, user_domains('todo'))
-        else:
-            flash('You need to select a domain to access this page', 'error')
-            return redirect(url_for('SelectDomainView.list', next=request.url))
+        flash('You need to select a domain to access this page', 'error')
+        return redirect(url_for('SelectDomainView.list', next=request.url))
 
 
 def is_valid_user_domain(hq_domain):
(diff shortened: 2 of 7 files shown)


hq_superset/hq_domain.py
```diff
@@ -0,0 +1,48 @@
...
+	if not request.url_rule.endpoint in DOMAIN_REQUIRED_VIEWS:
+		return
+	else:
+		hq_domain = request.cookies.get('hq_domain')
+		if hq_domain and is_valid_user_domain(hq_domain):
+			import pdb; pdb.set_trace()
+			g.hq_domain = hq_domain
+			# add_domain_links(hq_domain, user_domains('todo'))
+		else:
+			flash('You need to select a domain to access this page', 'error')
```
proteusvacuum, line 34: Do you think it's reasonable that if the user only has a single domain they can access, then we should just set that right away? I think that might be the case for most non-superusers.

proteusvacuum (commented):
This is looking great so far!

hq_superset/hq_domain.py
sravfeyn: Good idea!

sravfeyn:
@proteusvacuum Akash, Norman and I talked a bit today about this. We decided to offer domain-switching only at login, so a user would have to logout to switch the domain. The reason is that, with cookie-based domain switching, the user won't be able to access two domains at the same time in two tabs anyway. And implementing this means a bit more extra work. We will still keep this cookie based code around, so that we can support embedding. We can also come back add this feature at a latter point, since we know we can do this.

proteusvacuum:
@sravfeyn that sounds reasonable, thanks!

hq_superset/hq_domain.py
```diff
@@ -0,0 +1,48 @@
+import jinja2, os
+from flask import request, g, flash, redirect, url_for
+
+DOMAIN_REQUIRED_VIEWS = [
+	'TableModelView.list'
```
kaapstorm, line 5:
```suggestion
    'TableModelView.list'
```
This module sometimes uses spaces to indent, but mostly tabs. Please use four-space indents.

hq_superset/hq_domain.py
```diff
@@ -51,8 +58,11 @@ def is_valid_user_domain(hq_domain):
 
 
 def user_domains(user):
-    # Todo Implement based on persisted user domain membership data
-    return ['a', 'b']
+    # This should be set by oauth_user_info after OAuth
+    return [
+        d["domain_name"]
+        for d in session["user_hq_domains"]["objects"]
```
sravfeyn, line 64: @proteusvacuum @akashkj @kaapstorm 
I have modified the select-domain methodology a bit. The user's superset enabled domains are stored in the session at login. Immediately after the login, the user gets redirected to a select domain page (unless there is only a single domain), the selected domain is set on the request cookie for the rest of the session. To select another domain, user would have to navigate to the select domain page or logout/login.

sravfeyn:
@proteusvacuum  and @kaapstorm I ended up doing the role creation and syncing part as well in this PR. See below commit. 

https://github.com/dimagi/hq_superset/pull/2/commits/e6b02aab2977d7d017ea83c0e0d366b11a3bf23f

hq_superset/oauth.py
```diff
@@ -12,6 +12,8 @@ def oauth_user_info(self, provider, response=None):
         if provider == 'commcare':
             logging.debug("Getting user info from {}".format(provider))
             user = self.appbuilder.sm.oauth_remotes[provider].get("api/v0.5/identity/", token=response).json()
+            domains = self.appbuilder.sm.oauth_remotes[provider].get("api/v0.5/user_domains/", token=response).json()
+            session["user_hq_domains"] = domains
```
akashkj, line 17: Can store the key in a constant, to re-use.

hq_superset/hq_domain.py
akashkj: Is there possibility of getting null error when there is no valid domain for the user?

hq_superset/hq_domain.py
```diff
@@ -0,0 +1,72 @@
...
+
+def ensure_domain_selected():
+    # Check if a hq_domain cookie is set
+    #   Ensure necessary roles, permissions and DB schemas are created for the domain
+    import superset
+    if request.url_rule.endpoint in DOMAIN_EXCLUDED_VIEWS:
+        return
+    hq_domain = request.cookies.get('hq_domain')
+    valid_domains = user_domains(current_user)
+    # Todo; Handle no superset enabled domains case
```
proteusvacuum, line 46: What's the plan for this? Just redirect to a 404?

hq_superset/hq_domain.py
```diff
@@ -0,0 +1,72 @@
...
+    if hq_domain in valid_domains:
+        g.hq_domain = hq_domain
+    elif len(valid_domains) == 1:
+        g.hq_domain = valid_domains[0]
```
proteusvacuum, line 50: 👍🏽

hq_superset/hq_domain.py
proteusvacuum: I like that approach

hq_superset/oauth.py
```diff
@@ -33,6 +36,53 @@ def set_oauth_session(self, provider, oauth_response):
...
+        # This creates DB schema, role and schema permissions for the domain and
+        #   assigns the role to the current_user
+        # self.create_schema(domain)
+        permission = self.ensure_schema_perm_created(domain)
+        role = self.ensure_domain_role_created(domain)
+        role.permissions = [permission]
+        # Todo; may need to add the permission to admin role as well
+        self.get_session.add(role)
+        self.get_session.commit()
+        import pdb; pdb.set_trace()
```
proteusvacuum, line 60: eep

hq_superset/oauth.py
```diff
@@ -33,6 +36,53 @@ def set_oauth_session(self, provider, oauth_response):
...
+        current_user.roles = filtered_roles + [role]
+        self.get_session.add(current_user)
+        self.get_session.commit()
+
+
+def get_ucr_database():
+    from superset import db
+    from superset.models.core import Database
+    # Todo; get actual DB once that's implemented
+    return db.session.query(Database).filter_by(database_name="HQ Data").one()
```
proteusvacuum, line 77: Is there a way we can add this in a config file somewhere? Even if it is something like `settings` for django.

hq_superset/__init__.py
```diff
@@ -8,4 +8,11 @@ def patch_superset_config(config):
 def flask_app_mutator(app):
     # Import the views (which assumes the app is initialized) here
     # return
-    pass
+    from . import hq_domain
+    from . import views
+    from superset.extensions import appbuilder
+    appbuilder.add_view(views.SelectDomainView, 'Select Domain')
+    app.before_request_funcs.setdefault(None, []).append(
```
proteusvacuum, line 15: I think you should be using the [`@before_request`](https://flask.palletsprojects.com/en/2.1.x/api/#flask.Flask.before_request) decorator rather than mutating this list here: 
https://flask.palletsprojects.com/en/2.1.x/api/#flask.Flask.before_request_funcs

>To register a function, use the [before_request()](https://flask.palletsprojects.com/en/2.1.x/api/#flask.Flask.before_request) decorator.
> 
> This data structure is internal. It should not be modified directly and its format may change at any time.

sravfeyn: Using the decorator doesn't work in the case of superset/FAB, because the way the superset makes the app object available. The app object needs to be available when using it as decorator. That's the reason, I had to do it this way.

hq_superset/oauth.py
sravfeyn: Yeah, that's a TODO for future, I think we can add it in the superset config file.

hq_superset/hq_domain.py
sravfeyn: Actually, I removed this comment, the behaviour is that it keeps going to select-domain page where no domains are available to select.

hq_superset/oauth.py
sravfeyn: Indeed.

sravfeyn:
Please see the full PR https://github.com/dimagi/hq_superset/pull/7/files

hq_superset/__init__.py
proteusvacuum: Instead of this line, can you just do this? 

```suggestion
    app.before_request(hq_domain.before_request_hook)
```

hq_superset/oauth.py
sravfeyn: This is removed in the other PR https://github.com/dimagi/hq_superset/pull/7/files#

hq_superset/__init__.py
sravfeyn: I think this means that any other hooks elsewhere will get removed, no?

proteusvacuum: I don't think so. This would be the same as decorating the function directly. Using `before_request` will do exactly what you have done here, but in the "supported way".

https://github.com/pallets/flask/blob/bd56d19b167822a9a23e2e9e2a07ccccc36baa8d/src/flask/scaffold.py#L555-L556

PR #5, opened 2022-04-25
Add custom roles with schema specific permission

akashkj:
To create custom role and its related permission, if it does not exist.

### DIFF
--- a/hq_superset/oauth.py
+++ b/hq_superset/oauth.py
@@ -1,5 +1,7 @@
 import logging
+
 from superset.security import SupersetSecurityManager
+
 
 class CommCareSecurityManager(SupersetSecurityManager):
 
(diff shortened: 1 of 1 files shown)


hq_superset/oauth.py
```diff
@@ -10,3 +12,33 @@ def oauth_user_info(self, provider, response=None):
...
+        for domain_name in domains.items():
+            role_created = self.create_custom_role(domain_name)
+            if not role_created:
+                continue
+            self.create_schema(domain_name)
+            pv = self.create_custom_permission(domain_name)
+            sesh = self.get_session
+            role_created.permissions = pv
+            sesh.merge(role_created)
+            sesh.commit()
```
sravfeyn, line 26: What happens if there is an exception, should there be some sort of rollback?

hq_superset/oauth.py
```diff
@@ -10,3 +12,33 @@ def oauth_user_info(self, provider, response=None):
...
+
+    def create_custom_permission(self, domain_name):
+        schema_access_perm_name = "schema_access"
+        schema_access_view_menu_name = "[{}].[{}]".format("superset", domain_name)
```
sravfeyn, line 30: What is this permission for?

sravfeyn:
Looking good, Akash. I made few comments. I understand that these are the utils, but there is no code that's calling these yet?

sravfeyn:
I ended up doing this myself here https://github.com/dimagi/hq_superset/pull/2/commits/e6b02aab2977d7d017ea83c0e0d366b11a3bf23f

akashkj:
Is this PR not required now then? @sravfeyn

hq_superset/oauth.py
akashkj: Format for schema access permission is <db>.<schema name>

sravfeyn:
Yeah, I think we can close this

akashkj:
Ref: https://github.com/dimagi/hq_superset/pull/7

PR #6, opened 2022-04-27
Add a view to refresh HQ UCR based dataset

sravfeyn:
This PR adds a view that that takes the domain and config_id via the URL and re-pulls the datasource when called. Below are the implementation details. I was able to test this locally with success.

- The domain and config_id can be passed via the URL (for e.g. http://localhost:8088/a/test-domain/datasource/update/49f40d61cadcc69ede285982f4001e82). 
- The dataset is replaced on each refresh.

There is few more todos in here, but I wanted to get this out to get any initial feedback. The todos are

- Assign a read permission on the created/updated dataset to the user's domain specific role.
- Create a schema for the domain if it doesn't exist
- Make date parsing work.
- Let the view return the url or the DB model ID of the dataset.
- Unit tests.

Note that the Oauth part is in a separate PR as well https://github.com/dimagi/hq_superset/pull/4

@kaapstorm @akashkj @proteusvacuum

### DIFF
--- a/hq_superset/utils.py
+++ b/hq_superset/utils.py
@@ -1,0 +1,3 @@
+
+def get_datasource_export_url(domain, datasource_id):
+    return f"a/{domain}/configurable_reports/data_sources/export/{datasource_id}?format=csv"
--- a/hq_superset/utils.py
+++ b/hq_superset/utils.py
@@ -1,3 +1,17 @@
+import sqlalchemy
 
 def get_datasource_export_url(domain, datasource_id):
     return f"a/{domain}/configurable_reports/data_sources/export/{datasource_id}?format=csv"
+
+
+def get_ucr_database():
+    # Todo; get actual DB once that's implemented
+    return db.session.query(Database).filter_by(database_name="HQ Data").one()
+
+
+def create_schema_if_not_exists(schema_name):
+    # Create a schema in the database where HQ's UCR data is stored
+    database = get_ucr_database()
+    engine = database.get_sqla_engine()
+    if not engine.dialect.has_schema(engine, schema_name):
+        engine.execute(sqlalchemy.schema.CreateSchema(schema_name))
--- a/hq_superset/views.py
+++ b/hq_superset/views.py
@@ -1,0 +1,128 @@
+import pandas as pd
+import superset
+from flask import url_for, render_template, redirect, request, session, g
+from flask_appbuilder import expose, BaseView
+from flask_appbuilder.security.decorators import has_access, permission_name
+from flask_login import current_user
+from io import BytesIO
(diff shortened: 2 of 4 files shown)


hq_superset/utils.py
```diff
@@ -1,3 +1,17 @@
...
 def get_datasource_export_url(domain, datasource_id):
     return f"a/{domain}/configurable_reports/data_sources/export/{datasource_id}?format=csv"
+
+
+def get_ucr_database():
+    # Todo; get actual DB once that's implemented
+    return db.session.query(Database).filter_by(database_name="HQ Data").one()
+
+
+def create_schema_if_not_exists(schema_name):
```
sravfeyn, line 12: FYI @akashkj @proteusvacuum @kaapstorm   I think we can avoid https://github.com/dimagi/hq_superset/pull/1, let me know if this is missing anything from that PR.

proteusvacuum: LGTM

hq_superset/views.py
```diff
@@ -0,0 +1,122 @@
...
+        from .oauth import get_valid_cchq_oauth_token
+        res = refresh_hq_datasource(domain, datasource_id)
+        return res
+
+
+class CCHQApiException(Exception):
+    pass
+
+
+def refresh_hq_datasource(domain, datasource_id):
```
kaapstorm, line 35: Just a thought: This might be easier to write tests for if it's split up into smaller chunks. Maybe something like ...
```python
def refresh_hq_datasource(domain, datasource_id):
    zipfile = fetch_zipfile(domain, datasource_id)
    data_frame = read_csv(zipfile)
    upload_to_table(data_frame)
    return "success"
```

kaapstorm (commented):
This is looking good so far. Just a couple of thoughts.

hq_superset/views.py
sravfeyn: Yeah, definitely, there is going to be good amount of refactor once we start writing tests.

hq_superset/views.py
```diff
@@ -0,0 +1,122 @@
...
+    # This method pulls the data from CommCareHQ and creates/replaces the
+    #   corresponding Superset dataset
+    datasource_url = get_datasource_export_url(domain, datasource_id)
+    provider = superset.appbuilder.sm.oauth_remotes["commcare"]
+    oauth_token = get_valid_cchq_oauth_token()
```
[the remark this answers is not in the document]
sravfeyn: Since HQ is the only provider, we don't need to make that generic?

sravfeyn:
Please see the full PR https://github.com/dimagi/hq_superset/pull/7/files

PR #8, opened 2022-05-23
Handle failures at HQ end

sravfeyn:
@akashkj   @kaapstorm  @proteusvacuum

### DIFF
--- a/hq_superset/oauth.py
+++ b/hq_superset/oauth.py
@@ -1,7 +1,7 @@
 import logging
 import superset
 import time
-from flask import session
+from flask import flash, session
 from flask_login import current_user
 from requests.exceptions import HTTPError
 from superset.security import SupersetSecurityManager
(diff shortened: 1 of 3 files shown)


hq_superset/oauth.py
```diff
@@ -10,18 +10,33 @@
...
-            domains = self.appbuilder.sm.oauth_remotes[provider].get("api/v0.5/user_domains?feature_flag=superset-analytics", token=response).json()
+            logger.debug("Getting user info from {}".format(provider))
+            user = self._get_hq_response("api/v0.5/identity/", provider, response)
+            domains = self._get_hq_response("api/v0.5/user_domains?feature_flag=superset-analytics", provider, response)
             session[SESSION_USER_DOMAINS_KEY] = domains
-            logging.debug("user - {}".format(user))
+            logger.debug("user - {}".format(user))
             return user
 
+    def _get_hq_response(self, endpoint, provider, token):
```
kaapstorm, line 29: Is `token` here, `response` on line 19, and `oauth_response` on line 40, all the same data type? If so, it would be cool if they all had the same name.

(... Alternatively if using different names for the same data type seems best in this situation, you could use a [TypedDict](https://peps.python.org/pep-0589/) instead of the sample on line 46, to indicate that they're all the same thing. e.g.
```python
OAuthResponse = TypedDict('OAuthResponse', {
    'access_token': str,  # e.g. 'AcLBaXPC7HvJiefUYECBWOd4rCN6L9'
    'expires_in': int,  # e.g. 900
    'token_type': str,  # e.g. 'Bearer'
    'scope': str,  # e.g. 'view', 'reports:view access_apis'
    'refresh_token': str,  # e.g. 'kXU2Xo4RLn1UCYJMX2KWaic1kxI0PP'
    'expires_at': int,  # e.g. 1650872906
})
```
...
```python
    def _get_hq_response(self, endpoint, provider, token: OAuthResponse):
```
(This suggestion ignores the style guide we drew up when discussing type hints in HQ. So do the Superset and Flask AppBuilder code bases. We should do whatever maximises the understanding of future devs.))

kaapstorm (approved):
Looks good. Just a non-blocking suggestion about variable names.

hq_superset/oauth.py
sravfeyn: Agreed about the names! The schema is different sometimes, so that might not always be the case. I will update the names nevertheless in a separate PR.

PR #13, opened 2022-07-21
Column data types

kaapstorm:
Determines column data types for imported datasets.

Probably easier to review by commit, cos the first 3 commits don't do anything.


hq_superset/utils.py
```diff
@@ -41,3 +45,26 @@ def get_role_name_for_domain(domain):
...
+    # TODO: How are array indicators handled in CSV export?
+
+    COL_ATTR = 'column_id'
+    pandas_dtypes = {
+        'date': 'datetime64[ns]',
+        'datetime': 'datetime64[ns]',
+        'string': 'string',
+        'integer': 'int64',
+        'decimal': 'float64',
+        'small_integer': 'int8',  # TODO: Is this true?
```
Charl1996, line 65: According to this IBM document a small_integer is 2 bytes, although I'm not sure if the Dimagi dev who decided to call it "small_integer" followed this standard. I wonder what data will be stored in the column with this type?

I see in HQ it's used [here](https://github.com/dimagi/commcare-hq/blob/ea1f477742ada5ba7898c21522dac4e71f70e779/corehq/apps/aggregate_ucrs/column_specs.py#L338), so I'm assuming that this datatype will contain a 1 or a 0?

hq_superset/views.py
```diff
@@ -114,17 +115,16 @@ def refresh_hq_datasource(domain, datasource_id, display_name):
...
             df = pd.concat(
                 pd.read_csv(
                     chunksize=1000,
                     filepath_or_buffer=csv_file,
                     encoding="utf-8",
-                    # Todo; make date parsing work
-                    parse_dates=True,
-                    infer_datetime_format=True,
+                    parse_dates=date_columns,
+                    date_parser=parse_date,
```
Charl1996, line 127: So pandas will send the `date_columns` through the `parse_date` function when it's trying to read the data into the Dataframe?

Charl1996 (approved):

hq_superset/utils.py
```diff
@@ -37,3 +47,51 @@ def get_role_name_for_domain(domain):
...
+    See corehq/apps/userreports/datatypes.py for possible data types.
+    """
+    # TODO: How are array indicators handled in CSV export?
+    pandas_dtypes = {
+        'date': 'datetime64[ns]',
+        'datetime': 'datetime64[ns]',
+        'string': 'string',
+        'integer': 'int64',
+        'decimal': 'float64',
+        'small_integer': 'int8',  # TODO: Is this true?
```
sravfeyn, line 65: From my understanding, this maps to SQLAlchemy's [SmallInteger](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/sql/columns.py#L52), which SQLAlchemy [maps](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.SmallInteger) to Postgres's SMALLINT which is of [2 bytes](https://www.postgresql.org/docs/current/datatype-numeric.html).

hq_superset/utils.py
```diff
@@ -37,3 +47,51 @@ def get_role_name_for_domain(domain):
...
+
+def get_column_dtypes(datasource_defn):
+    """
+    Maps UCR column data types to Pandas data types.
+
+    See corehq/apps/userreports/datatypes.py for possible data types.
+    """
+    # TODO: How are array indicators handled in CSV export?
+    pandas_dtypes = {
+        'date': 'datetime64[ns]',
```
sravfeyn, line 60: I am curious if where the format of the values here come from. Per pandas docs [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html) this should be SQLAlchemy datatype?

hq_superset/utils.py
```diff
@@ -37,3 +47,51 @@ def get_role_name_for_domain(domain):
...
+        'datetime': 'datetime64[ns]',
+        'string': 'string',
+        'integer': 'int64',
+        'decimal': 'float64',
+        'small_integer': 'int8',  # TODO: Is this true?
+    }
+    column_dtypes = {'doc_id': 'string'}
```
sravfeyn, line 67: I think there is UUID field in Postgres? Though, I am not sure if Sqlalchemy/pandas has support for that. 

The main thing to be concerned about though is if the column will be mapped to be a bounded length string or unbounded length string. See https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Unicode which can be bounded.

hq_superset/views.py
```diff
@@ -99,34 +105,30 @@ class CCHQApiException(Exception):
...
-        df = pd.concat(
-            pd.read_csv(
-                chunksize=1000,
-                filepath_or_buffer=zipfile.open(filename),
-                encoding="utf-8",
-                # Todo; make date parsing work
-                parse_dates=True,
-                infer_datetime_format=True,
-                keep_default_na=True,
+        with get_csv_file(provider, token, domain, datasource_id) as csv_file:
```
sravfeyn, line 120: Nitpicking here (feel free to ignore), the context process needs to be just for the ZipFile.

hq_superset/utils.py
```diff
@@ -37,3 +47,51 @@ def get_role_name_for_domain(domain):
...
+def get_column_dtypes(datasource_defn):
+    """
+    Maps UCR column data types to Pandas data types.
+
+    See corehq/apps/userreports/datatypes.py for possible data types.
+    """
+    # TODO: How are array indicators handled in CSV export?
+    pandas_dtypes = {
+        'date': 'datetime64[ns]',
+        'datetime': 'datetime64[ns]',
```
sravfeyn, line 61: Does this keep the timezone info intact in the final Postgres DB?

sravfeyn:
This looks great! Thank you!

hq_superset/views.py
kaapstorm: Exactly. Pandas comes with a [`pd.to_datetime()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_datetime.html) function you can use, but it's big and caters for a lot of different formats. We know that all the dates and datetimes from HQ are going to be ISO-8601 format, so we can just use Python's `fromisoformat()`.

hq_superset/utils.py
kaapstorm: These aren't SQLAlchemy datatypes. They are Pandas [Dataframe datatypes](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.dtypes.html), which we use [here, when reading the CSV](https://github.com/dimagi/hq_superset/pull/13/commits/63c22451dbb446500f325da37a65e5013eae4d23#diff-904cccd73cc4407182ae5580289ad118ed1be682677edf1c7e0ba03489788699R129).

If "small_integer" in UCRs is 2 bytes (-32,768 - 32,767) we should be fine. (I suspect we use it for booleans as well, because `'boolean'` [is not included in DATA_TYPE_CHOICES](https://github.com/dimagi/commcare-hq/blob/cc93bc6ab718bde08e8d16e374d5cc25eff94958/corehq/apps/userreports/datatypes.py#L9). (?!) ) I just don't want to be in a situation where 8 bytes is not enough.

hq_superset/utils.py
kaapstorm: These values are being passed to [`read_csv()`](https://github.com/dimagi/hq_superset/pull/13/commits/63c22451dbb446500f325da37a65e5013eae4d23#diff-904cccd73cc4407182ae5580289ad118ed1be682677edf1c7e0ba03489788699R129), when the Dataframe is being created, not to `to_sql()`. `to_sql()` respects the Dataframe's data types.

hq_superset/utils.py
kaapstorm: There is a UUID field in Postgres, but [SQLAlchemy](https://docs.sqlalchemy.org/en/14/core/type_basics.html) doesn't have a UUID data type.

Good call on the bounded/unbounded strings! I will check what type string columns end up with in Postgres.

hq_superset/views.py
kaapstorm: You are right that the context processor is for `ZipFile`. What I was thinking was that I could move more code out of the `refresh_hq_datasource()` function, and using a context process closes both the `ZipFile`, and the CSV file inside it. My primary objective was to move the code for fetching the CSV file into its own function. The context processor was just convenient.

hq_superset/utils.py
kaapstorm: The "Specify date/datetime columns" commit actually might ... by accident!

When I was writing this, I got it into my head that Dataframes don't support timezones, [but that's not true](https://pandas.pydata.org/pandas-docs/stable/user_guide/basics.html#attributes-and-underlying-data): "NumPy doesn’t have a dtype to represent timezone-aware datetimes" not Pandas.

I'll test timezone-aware UCR datetimes, and change the spelling of these so it's less misleading that these are Pandas data types. (You'll see in the "Specify date/datetime columns" commit that we don't specify data types for dates and datetimes.)

PR #17, opened 2022-11-11
Add additional roles on signup

Charl1996:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-2363)

Add a constant specifying additional roles for a user on top of the base "Gamma" role.


superset_config.py
```diff
@@ -29,4 +29,15 @@
...
+AUTH_USER_REGISTRATION_ROLE = "Gamma"
+
+# Any other additional roles to be assigned to the user on top of the base role
+# Note: by design we cannot use AUTH_USER_REGISTRATION_ROLE to
+# specify more than one role
+AUTH_USER_ADDITIONAL_ROLES = ["sql_lab"]
+
+try:
+    # Overwrite repo settings with local settings
+    from local_settings import *
```
sravfeyn, line 41: Good idea, could name it as local_config to keep it consistent with flask/superset terminology.

superset_config.py
```diff
@@ -29,4 +29,15 @@
...
 
 # The default user self registration role
-AUTH_USER_REGISTRATION_ROLE = "Gamma"
\ No newline at end of file
+AUTH_USER_REGISTRATION_ROLE = "Gamma"
+
+# Any other additional roles to be assigned to the user on top of the base role
+# Note: by design we cannot use AUTH_USER_REGISTRATION_ROLE to
+# specify more than one role
+AUTH_USER_ADDITIONAL_ROLES = ["sql_lab"]
```
sravfeyn, line 37: Is `Gamma` not a subset of `sql_lab`?

sravfeyn: I don't know as well, but the behaviour seemed as if it's so. You can find that by comparing the permissions for each role under Edit Role where role permissions are listed in your local superset env.

SmittieC: https://superset.apache.org/docs/security/#:~:text=The%20sql_lab%20role%20grants%20access%20to%20SQL%20Lab.%20Note%20that%20while%20Admin%20users%20have%20access%20to%20all%20databases%20by%20default%2C%20both%20Alpha%20and%20Gamma%20users%20need%20to%20be%20given%20access%20on%20a%20per%20database%20basis.

superset_config.py
Charl1996: Agree!

superset_config.py
```diff
@@ -29,4 +29,15 @@
...
+
+try:
+    # Overwrite repo settings with local settings
+    from local_config import *
+except:
```
SmittieC, line 42: I think we should catch `ModuleNotFound` explicitly here. Reason being that if there's some logic in the `local_config.py` module that is erroneous, it will be masked by this `except` clause, resulting in unexpected behaviour.

superset_config.py
sravfeyn: Great, thanks for digging.

sravfeyn:
Looks good now. This should be QAed on http://commcare-analytics-demo.dimagi.com/ before merging similar to https://dimagi-dev.atlassian.net/browse/QA-174 with any additional SQL lab related tests to make sure queries can't access other domains data.

Charl1996:
@sravfeyn I just bumped the superset version on this PR (probably a good idea?)

Charl1996:
Seems to work fine on the demo version.

sravfeyn:
>  just bumped the superset version on this PR 

I presume you mean the hq_superset version. If that's so, we don't really follow versioning, neither is this is a proper pip package. So there's really no point. 

But it does feel very nice to bump up the versions 😆

Charl1996:
@sravfeyn After testing again on the demo version it seems this works. Also, I can't access tables from other domains through SQL lab. Feel free to also verify if you like.

Could I perhaps get an approval if all is well?

sravfeyn (approved):
Great work, Charl, thank you!

PR #19, opened 2022-11-24
Handle array datatype columns in datasets

Charl1996:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-2208)

Handle datasource array datatype. Arrays will be stored in postgres as a collection of strings. I'm not sure if we can assume that it will always be strings?

As far as testing goes, I could pull the data from postgres through SQL Lab in superset and view it. Not sure if there is other tests I can also do?

### DIFF
--- a/hq_superset/views.py
+++ b/hq_superset/views.py
@@ -1,10 +1,13 @@
 import logging
+import ast
+import pandas as pd
+import superset
+import sqlalchemy
+
 from contextlib import contextmanager
 from io import BytesIO
 from zipfile import ZipFile
-
-import pandas as pd
-import superset
+from sqlalchemy.dialects import postgresql
 from flask import Response, abort, g, redirect, request
 from flask_appbuilder import expose
 from flask_appbuilder.security.decorators import has_access, permission_name
@@ -104,6 +107,13 @@ class CCHQApiException(Exception):
     pass
 
 
+def convert_to_array(string_array):
+    array_values = ast.literal_eval(string_array)
+    if array_values and array_values[0] is None:
+        return []
+    return array_values
+
+
 def refresh_hq_datasource(domain, datasource_id, display_name):
     """
     Pulls the data from CommCare HQ and creates/replaces the
@@ -115,7 +125,12 @@ def refresh_hq_datasource(domain, datasource_id, display_name):
     schema = get_schema_name_for_domain(domain)
     csv_table = Table(table=datasource_id, schema=schema)
     datasource_defn = get_datasource_defn(provider, token, domain, datasource_id)
-    column_dtypes, date_columns = get_column_dtypes(datasource_defn)
+    column_dtypes, date_columns, array_columns = get_column_dtypes(datasource_defn)
+
+    converters = {column: convert_to_array for column in array_columns}
+    # TODO: can we assume all array values will be of type TEXT?
+    sqlconverters = {column: postgresql.ARRAY(sqlalchemy.types.TEXT) for column in array_columns}
+
     try:
         with get_csv_file(provider, token, domain, datasource_id) as csv_file:
             df = pd.concat(
@@ -127,9 +142,9 @@ def refresh_hq_datasource(domain, datasource_id, display_name):
                     date_parser=parse_date,
                     keep_default_na=True,
                     dtype=column_dtypes,
+                    converters=converters,
                 )
             )
-
         database.db_engine_spec.df_to_sql(
             database,
             csv_table,
@@ -137,6 +152,7 @@ def refresh_hq_datasource(domain, datasource_id, display_name):
             to_sql_kwargs={
                 "chunksize": 1000,
                 "if_exists": "replace",
+                "dtype": sqlconverters,
             },
         )
 
@@ -147,10 +163,10 @@ def refresh_hq_datasource(domain, datasource_id, display_name):
         explore_database_id = database.explore_database_id
         if explore_database_id:
             expore_database = (
-                db.session.query(Database)
-                .filter_by(id=explore_database_id)
-                .one_or_none()
-                or database
+                    db.session.query(Database)
+                    .filter_by(id=explore_database_id)
+                    .one_or_none()
+                    or database
             )
 
         sqla_table = (
@@ -208,7 +224,6 @@ def get_datasource_defn(provider, oauth_token, domain, datasource_id):
 
 
 class SelectDomainView(BaseSupersetView):
-
     """
     Select a Domain view, all roles that have 'profile' access on 'core.Superset' view can access this
     """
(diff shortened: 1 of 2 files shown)


hq_superset/views.py
```diff
@@ -104,6 +107,13 @@ class CCHQApiException(Exception):
     pass
 
 
+def convert_to_array(string_array):
+    array_values = ast.literal_eval(string_array)
+    if array_values and array_values[0] is None:
```
SmittieC, line 112: I would either
1. give this check a descriptive name
2. extract this to a method which handle these kind of cornercases

hq_superset/views.py
```diff
@@ -115,7 +125,12 @@ def refresh_hq_datasource(domain, datasource_id, display_name):
     schema = get_schema_name_for_domain(domain)
     csv_table = Table(table=datasource_id, schema=schema)
     datasource_defn = get_datasource_defn(provider, token, domain, datasource_id)
-    column_dtypes, date_columns = get_column_dtypes(datasource_defn)
+    column_dtypes, date_columns, array_columns = get_column_dtypes(datasource_defn)
+
+    converters = {column: convert_to_array for column in array_columns}
```
SmittieC, line 130: Maybe use `column_name` instead of `column`? I.E

`converters = {column_name: convert_to_array for column_name in array_columns}`

hq_superset/views.py
```diff
@@ -104,6 +107,13 @@ class CCHQApiException(Exception):
     pass
 
 
+def convert_to_array(string_array):
```
kaapstorm, line 110: Might be nice to show what `string_array` is expected to look like. I love docstrings for doing this, but English also works.
```suggestion
def convert_to_array(string_array):
    """
    Converts the string representation of a list to a list.

    >>> convert_to_array("['hello', 'world']")
    ['hello', 'world']

    """
```

You could add unit tests to test the rest of the behavior ... or you could just add a few more doctests ...
```suggestion
def convert_to_array(string_array):
    """
    Converts the string representation of a list to a list.

    >>> convert_to_array("['hello', 'world']")
    ['hello', 'world']
    >>> convert_to_array("hello, world")
    Traceback (most recent call last):
      ...
    ValueError: malformed node or string: ...
    >>> convert_to_array("[None]")
    []
    >>> convert_to_array("'hello', 'world'")
    ('hello', 'world')

    """
```
I prefer to keep doctests short, and use unit tests to verify behavior, but I guess it depends where you find the balance between code readability and typing unit tests.

... Also, should we be catching that `ValueError`?

hq_superset/views.py
```diff
@@ -1,10 +1,13 @@
 import logging
+import ast
+import pandas as pd
+import superset
+import sqlalchemy
```
kaapstorm, line 5: :thinking: Importing pandas and sqlalchemy in a "views" module smells dodgy. I'm not a fan of modules called "utils" cos they often degrade into a dumping ground for code that belongs in other places, but at the moment `utils.py` is kind of a service layer, where pandas and sqlalchemy feel more at home.

kaapstorm: :thinking: :thinking: on second thoughts, this module is already doing HQ API client stuff, and database stuff. I still think we should move those to separate modules, but maybe not to bother in this PR.

kaapstorm (approved):

hq_superset/views.py
sravfeyn: Agreed Norman. On that note, I am the one that's mostly responsible for the not-so-good shape of the code here since this was written as an MVP, deprioritising unit tests and other standards. My idea was to refactor these things as part of Automated-tests task. If refactoring is difficult as part of that work, I am happy to create a separate ticket for refactor. Thanks for bringing this up!

hq_superset/views.py
Charl1996: Doctests are a great idea. Also, looking back at this function now I realize it's written a little narrow (ie. no error handling or so). I'll make some updates.

hq_superset/views.py
```diff
@@ -108,9 +108,40 @@ class CCHQApiException(Exception):
...
+    >>> convert_to_array("hello, world")
+    []
+    """
+
+    def array_is_falsy(array_values):
+        if not array_values:
+            return True
+        if array_values[0] is None:
+            return True
+        return False
```
kaapstorm, line 131: Careful:
```python
>>> array_is_falsy([None, 1, 2, 3])
True
```

Charl1996: Yes, I also thought of that scenario (and maybe should have included it in the docstrings to make it clear that it's expected behavior) and I still think it's true that it's a falsy array if there's a None at the start (or any other index...I should probably cater for that then too), since we really want an array with only real items to pass. Maybe I should rename the function to something like "array_contains_none_values", then it can become

```
def array_contains_none_values(array):
    if not array:
        return True
    if None in array:
        return None
    return False
``` 

That said, I'm open to suggestions.

Charl1996: I can't make up my mind about the previous comment, so I'll put it out here.

The motivation behind the function is to not store an array as `[None]` (as would be the case for datasource expressions resulting in None that gets exported to CSV as "[None]"), but rather `[]`. Now, I don't know if I should be making the assumption that, if an array contains _any_ None values, that that array is falsy. If I decide not to make that assumption I'd have to do something like
```
if len(array) == 1 and array[0] is None:
    return True
``` 

Maybe that's better?

Charl1996: Actually, now that I think of it I prefer to allow for None values being nested between real values, since I don't think I should be making assumptions about how array type datasource indicator expressions are being evaluated in HQ, but I know [None] should be [].

SmittieC: Is it acceptable for a `None` value to be stored? Would it be acceptable to just remove any `None` value and return the new array?

kaapstorm: > The motivation behind the function is to not store an array as [None] (as would be the case for datasource expressions resulting in None that gets exported to CSV as "[None]")

I think this is crucial context. It makes sense that `[]` and `[None]` should be treated as an empty list.

In that case, what about
```python
def array_is_falsy(array_values):
    """
    Returns ``True`` if ``array_values`` is empty or ``[None]``, which
    occurs when ``None`` is exported to CSV. Otherwise returns
    ``False``.

    >>> array_is_falsy([None])
    True

    """
    return not array_values or array_values == [None]
```

kaapstorm: (Sorry. I should have kept scrolling. I didn't notice the commit you added 7 hours ago. :facepalm: )

Charl1996: I like the one-liner!

SmittieC (approved):

PR #23, opened 2023-03-01
Process chunks separately

sravfeyn:
@SmittieC I am going to try this to see if it fixes the 504 for large CSVs https://dimagi-dev.atlassian.net/browse/SC-2596


SmittieC (approved):
Looks good to me. I see there's a test failing though

PR #24, opened 2023-04-27
Import larger UCRs via celery

sravfeyn:
https://dimagi-dev.atlassian.net/browse/SC-2612


hq_superset/views.py
```diff
@@ -140,26 +149,58 @@ def array_is_falsy(array_values):
 
     return array_values
 
+# ~5MB
+ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES = 5000000
```
SmittieC, line 153: Can this be moved to the top of the file or the superset config file, then we can read it from there?

hq_superset/views.py
```diff
@@ -140,26 +149,58 @@ def array_is_falsy(array_values):
...
+        response = refresh_hq_datasource(
+            domain, datasource_id, display_name, path, datasource_defn, None
+            )
+        os.remove(path)
+        return response
+    else:
+        limit_in_mb = int(ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES / 1000000)
+        flash(f"The datasource is being refreshed in the background as it is"
+              " larger than {limit_in_mb} MB. This may take a while, please wait for it to finish",
+              "info")
```
SmittieC, line 176: Totally open for dispute, but I feel like the `AsyncImportHelper` should encaptulate the `queue_refresh_task`, so that the view doesn't have to worry about this. I think it might be better to use the `AsyncImportHelper` as a "service" that handles all the bells and whistles around async tasks for the caller. Thoughts?

hq_superset/views.py
```diff
@@ -230,17 +275,24 @@ def refresh_hq_datasource(domain, datasource_id, display_name):
 
 
 @contextmanager
-def get_csv_file(provider, oauth_token, domain, datasource_id):
+def get_datasource_file(path):
```
SmittieC, line 278: I smell a DataSourceHelper/DataSourceService brewing...In the meantime to satisfy the hunger, I'd suggest this function to live in the `utils.py` file rather to keep the views clean?

hq_superset/views.py
sravfeyn: If you look at `queue_refresh_task`, its just three lines, so there isn't a ton of code to move off of it to `AsyncImportHelper`. `AsyncImportHelper` is really just a group of related functions. I would hold off making abstractions until we really need to add a new task.

hq_superset/views.py
sravfeyn: >  I'd suggest that this function should live in the utils.py file rather to keep the views clean?

Agreed, I will move it.

hq_superset/views.py
sravfeyn: Sure. Let me move this to `utils.py` though, instead of the config file. Having it in a config file is required only if we need to customize it per env.

SmittieC (approved):
LGTM :rocket: Not sure if we want someone else to also have a glance at this

Charl1996 (approved):
Nothing major from me.

hq_superset/views.py
```diff
@@ -141,17 +152,47 @@ def array_is_falsy(array_values):
...
+    path, size = download_datasource(provider, token, domain, datasource_id)
+    datasource_defn = get_datasource_defn(provider, token, domain, datasource_id)
+    if size < ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES:
+        response = refresh_hq_datasource(
+            domain, datasource_id, display_name, path, datasource_defn, None
+            )
+        os.remove(path)
+        return response
+    else:
+        limit_in_mb = int(ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES / 1000000)
```
kaapstorm, line 173: Not worth the hassle of changing, but useful to know: Python 3.6+ supports [using "_" as a thousands separator](https://peps.python.org/pep-0515/) to make large numbers more readable, like
```python
>>> foo = 10_000_000
>>> foo
10000000
```
(Also works for hex (`0xCAFE_F00D`) and binary (`0b_0011_1111_0100_1110`).)

kaapstorm (approved):

hq_superset/views.py
sravfeyn: That's neat!

PR #25, opened 2023-05-15
Pin conflicting versions

sravfeyn:


mkangia:
Hey @sravfeyn @kaapstorm 
is this still needed?

sravfeyn:
I think so (?), it can be verified by looking the versions up on the server

kaapstorm:
I think once we have followed [the steps to patch apache-superset 3.1.0](https://github.com/dimagi/hq_superset/blob/master/apache-superset.md) to build a dimagi-superset for 3.1.0, then we can test whether we still need to pin those versions. I don't think there is anything in Dimagi's changes that specifically require those versions. (Is there @sravfeyn ?)

PR #26, opened 2023-07-26
Add permission to save datasource for domain roles

zandre-eng:
Link to ticket [here](https://dimagi-dev.atlassian.net/browse/SC-2708).

This change adds the `("can_save", "Datasource")` permission for domain roles when calling `sync_domain_role()`. Testing has been done to ensure that this new permission does not allow users to access datasources that they do not have access to.

### DIFF
--- a/hq_superset/utils.py
+++ b/hq_superset/utils.py
@@ -157,6 +157,12 @@ class DomainSyncUtil:
         if not engine.dialect.has_schema(engine, schema_name):
             engine.execute(sqlalchemy.schema.CreateSchema(schema_name))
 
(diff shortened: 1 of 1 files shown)


sravfeyn:
Thanks for testing @zandre-eng! Looks good.

Should we add an automated test for the same manual test you did, something similar to [test_non_user_domain_cant_be_selected](https://github.com/dimagi/hq_superset/blob/master/hq_superset/tests/test_views.py#L190)?

zandre-eng:
@sravfeyn  After our offline discussion to try and avoid requiring this permission, I have confirmed that this is not a priority and will be closing off this branch.

PR #29, opened 2023-11-22
Webhook

kaapstorm:
This is the start on the API endpoint for making changes to datasets.

The first five commits are only cleanup, and a small refactor to move a function into `utils.py` so that both `tasks` and `views` import from `utils`.

[Create webhook endpoint](https://github.com/dimagi/hq_superset/pull/29/commits/d77b480f64ed22b9e2861973d2e0421bb8f6f689) is the first useful commit. It defines a dataclass for the request payload, and adds the scaffolding for a view for the endpoint.

fyi @Charl1996 @SmittieC


hq_superset/views.py
```diff
@@ -236,3 +240,48 @@ def select(self, hq_domain):
...
+        try:
+            request_json = json.loads(request.get_data(as_text=True))
+        except json.JSONDecodeError:
+            return json_error_response(
+                'Invalid JSON syntax',
+                status=HTTPStatus.BAD_REQUEST.value,
+            )
+
+        try:
+            change = DataSetChange(**request_json)
```
SmittieC, line 275: Early suggestion: What if you move lines 275 and 284-287 into the first try on line 266 (so the happy case is inside the try) and then handle the "sad" case in the rest of the function?

hq_superset/models.py
```diff
@@ -0,0 +1,97 @@
...
+
+from datetime import datetime
+from superset import db
+from authlib.integrations.sqla_oauth2 import OAuth2ClientMixin
+from werkzeug.security import check_password_hash, generate_password_hash
+
+
+@dataclass
+class DataSetChange:
+    action: Literal['upsert', 'delete']
```
SmittieC, line 15: This is not needed anymore

hq_superset/api.py
```diff
@@ -0,0 +1,126 @@
...
+        if request.content_length > self.MAX_REQUEST_LENGTH:
+            return json_error_response(
+                HTTPStatus.REQUEST_ENTITY_TOO_LARGE.description,
+                status=HTTPStatus.REQUEST_ENTITY_TOO_LARGE.value,
+            )
+
+        try:
+            request_json = json.loads(request.get_data(as_text=True))
+            change = DataSetChange(**request_json)
+            update_dataset(change)
```
SmittieC, line 109: Nitty: Thoughts on moving `update_dataset` into `DataSetChange` and chaning the name to `DataSet` so that that object represents a dataset, much like django's models represent a Row?

hq_superset/models.py
```diff
@@ -0,0 +1,97 @@
...
+    data_source_id: str
+    data: Dict[str, Any]
+
+    def __post_init__(self):
+        if 'doc_id' not in self.data:
+            raise TypeError("'data' missing required key: 'doc_id'")
+
+
+class HQClient(db.Model, OAuth2ClientMixin):
```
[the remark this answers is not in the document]
SmittieC: Thought: We use a separate migration branch for our migrations. We must then somehow keep these separate from superset migrations, so we'll need a way of extracting the generated migration file.....or...we could manually create our migration files...

README.md
```diff
@@ -117,6 +117,41 @@ code you want to test will need to be in a module whose dependencies
...
+3. Edit `env.py` and comment out the following lines:
+   ```
+   hq_data_uri = current_app.config['SQLALCHEMY_BINDS'][HQ_DATA]
+   decoded_uri = urllib.parse.unquote(hq_data_uri)
+   config.set_main_option('sqlalchemy.url', decoded_uri)
+   ```
+
+Those changes will allow Alembic to connect to the "HD Data" database
+without the need to instantiate Superset's Flask app. You can now
+autogenerate your new table with:
```
SmittieC, line 147: :partying_face:

hq_superset/tasks.py
```diff
@@ -16,34 +13,3 @@ def refresh_hq_datasource_task(domain, datasource_id, display_name, export_path,
         AsyncImportHelper(domain, datasource_id).mark_as_complete()
         raise
     os.remove(export_path)
-
```
SmittieC: From the commit message:
> so that it can get the OAuth token to authenticate with HQ.

Speaking out of inexperience in this area, but can't we simply pass the token to the celery task?

kaapstorm:
Closing. This branch has been rebased and can be reviewed at PR https://github.com/dimagi/hq_superset/pull/41

hq_superset/tasks.py
kaapstorm: (Just closing the loop here. It needs the request session. We could serialize it and pass it to the task, but something about that feels awkward to me. I'm happy to be educated though. :) )

PR #30, opened 2023-12-06
Cs/sc 3069 subscribe to ds changes

Charl1996:
Opening as draft since I'm busy checking how we can better manage migrations in this repo.

The changes introduced in this PR is as follows:
1. Add and invoke task to subscribe to a specific datasource on HQ
2. Adds a new API for handling OAuth with superset as the authorization server.
3. Add decorator (`require_oauth`) to check access token on `/hq_webhook/change/` endpoint.

### DIFF
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -90,7 +90,7 @@ class OAuth(BaseApi):
         try:
             response = authorization.create_token_response()
         except NoResultFound:
-            return jsonify({"error": "Invalid credentials"}), 401
+            return jsonify({"error": "Invalid client"}), 401
 
         if response.status_code >= 400:
             return response
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -1,0 +1,93 @@
+import json
+from datetime import datetime, timedelta
+from flask_appbuilder.api import expose, BaseApi
+from flask import jsonify
+from superset import db
+from authlib.integrations.flask_oauth2 import AuthorizationServer
+from authlib.oauth2.rfc6749 import grants
+from superset.extensions import appbuilder
+from werkzeug.security import check_password_hash
+from authlib.integrations.sqla_oauth2 import OAuth2ClientMixin
+
+ONE_DAY_SECONDS = 60*60*24
+app = appbuilder.app
+
+
+class HQClient(db.Model, OAuth2ClientMixin):
+    __tablename__ = 'hq_oauth_client'
+
+    domain = db.Column(db.String(255), primary_key=True)
+
+    def check_client_secret(self, client_secret):
+        return check_password_hash(self.client_secret, client_secret)
+
+    def revoke_tokens(self):
+        tokens = db.session.execute(
+            db.select(Token).filter_by(client_id=self.client_id, revoked=False)
+        ).all()
+        for token, in tokens:
+            token.revoked = True
+            db.session.add(token)
+        db.session.commit()
+
+
+class Token(db.Model):
+    __tablename__ = 'hq_oauth_token'
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    client_id = db.Column(db.String(40), nullable=False, index=True)
+    token_type = db.Column(db.String(40))
+    access_token = db.Column(db.String(255), nullable=False, unique=True)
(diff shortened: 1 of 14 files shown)


hq_superset/models.py
```diff
@@ -0,0 +1,87 @@
...
+from superset import db
+from authlib.integrations.sqla_oauth2 import OAuth2ClientMixin
+from werkzeug.security import check_password_hash, generate_password_hash
+
+
+@dataclass
+class DataSetChange:
+    action: Literal['upsert', 'delete']
+    data_source_id: str
+    data: dict[str, Any]
```
SmittieC, line 15: I wonder if we might not want to have another dataclass for this `data` field. That way it will do its own validation and we will know exactly what keys we expect

hq_superset/utils.py
```diff
@@ -190,23 +197,236 @@ def get_datasource_file(path):
...
+
+    try:
+        array_values = ast.literal_eval(string_array)
+    except ValueError:
+        return []
+
+    if isinstance(array_values, tuple):
+        array_values = list(array_values)
+
+    # Test for corner cases
```
SmittieC, line 261: is this being tested?

Charl1996:
I hope I rebased correctly. It seems like it.

hq_superset/views.py
```diff
@@ -247,8 +232,8 @@ def post_dataset_change(self) -> FlaskResponse:
 
         try:
             request_json = json.loads(request.get_data(as_text=True))
-            change = DataSetChange(**request_json)  # raises TypeError
-            update_dataset(change)  # raises ValueError
+            change = DataSetChange(**request_json)
+            update_dataset(change)
```
Charl1996, line 236: @kaapstorm Should this be commented out?

.idea/.gitignore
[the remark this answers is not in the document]
Charl1996: Oh, I forgot to remove that (I did see it and forgot to remove it). Thanks for the suggestion about gitignore_global, I'll do that!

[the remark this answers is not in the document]
Charl1996: Removed

hq_superset/views.py
kaapstorm: No? That's the call that does the work.

hq_superset/api.py
```diff
@@ -0,0 +1,71 @@
...
+
+def query_client(client_id):
+    return HQClient.get_by_client_id(client_id)
+
+
+def save_token(token, request):
+    client = request.client
+    client.revoke_tokens()
+
+    expires_at = datetime.utcnow() + timedelta(seconds=ONE_DAY_SECONDS)
```
kaapstorm, line 27:
```suggestion
    expires_at = datetime.utcnow() + timedelta(days=1)
```

hq_superset/models.py
```diff
@@ -11,3 +17,71 @@ class DataSetChange:
...
+        return client.domain
+
+    def is_expired(self):
+        return datetime.utcnow() > self.expires_at
+
+    def is_revoked(self):
+        return self.revoked
+
+    def get_scope(self):
+        return self.scope
```
kaapstorm, line 87: It's not obvious to me why you created a method that returns a property that is already public, and it doesn't look like a subclass will override this method. Do we need it?

kaapstorm (approved):
All looking good. Just a few minor comments.

hq_superset/models.py
Charl1996: The `get_scope` is a method that the `require_oauth` decorator is looking for on the Token class, so it has to be defined as such. I'll add a comment and make that clear.

hq_superset/models.py
```diff
@@ -11,3 +17,71 @@ class DataSetChange:
...
+    def get_by_domain(cls, domain):
+        return db.session.query(HQClient).filter_by(domain=domain).first()
+
+    @classmethod
+    def get_by_client_id(cls, client_id):
+        return db.session.query(HQClient).filter_by(client_id=client_id).first()
+
+    @classmethod
+    def create_domain_client(cls, domain: str):
+        client_secret = str(uuid.uuid4())
```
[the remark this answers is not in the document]
Charl1996: Ah, thanks for the suggestion! Turns out uniqueness != cryptographically secure!

PR #31, opened 2024-01-24
Use `SUPERSET_CONFIG_PATH` env var to manage config

kaapstorm:
This change uses the `SUPERSET_CONFIG_PATH` environment variable to manage Superset configuration, both for tests and for normal operation.

Calling `hq_superset.patch_superset_config(superset_config)` is unnecessary. The same can be achieved by including 
```
FLASK_APP_MUTATOR = flask_app_mutator
CUSTOM_SECURITY_MANAGER = oauth.CommCareSecurityManager
```
in the config file.

I've added a comment to the start of `superset_config.example.py` to explain how to set the environment variable for normal operation. The environment variable is set automatically in `hq_superset/tests/base_test.py` for running tests.

### DIFF
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -4,12 +4,6 @@ import jinja2
 from flask import Blueprint
 
 
-def patch_superset_config(config):
-    from . import oauth
-    config.FLASK_APP_MUTATOR = flask_app_mutator
-    config.CUSTOM_SECURITY_MANAGER = oauth.CommCareSecurityManager
-
-
 def flask_app_mutator(app):
     # Import the views (which assumes the app is initialized) here
     # return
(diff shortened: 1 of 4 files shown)


SmittieC (approved):

SmittieC:
@kaapstorm When I upgrade typing extensions locally, it works, but the version is larger than that which superset wants, so maybe we should merge this branch and see what it does with the later supset version?

kaapstorm:
> When I upgrade typing extensions ... see what it does with the later supset version?

I've rebased off `master`, which should resolve this. I'm watching the tests.

kaapstorm:
Following up failures.

kaapstorm:
Failures were caused by `config_for_tests.py` diverging from GitHub workflow config.

PR #33, opened 2024-01-28
Upgrade to apache-superset==3.1.0

kaapstorm:
Upgrades hq_superset to apache-superset==3.1.0

This required a few tweaks:
* `superset.dataset.commands` moved to `superset.commands.dataset`
* `is_user_admin()` became `security_manager.is_admin()`
* `get_sqla_engine()` method became `get_sqla_engine_with_context()` context manager.
* The parameters of a couple of commands changed

That was about it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,3 @@
-"""
-Setup.py for superset-patchup
-"""
 from setuptools import setup, find_packages
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -5,31 +5,30 @@ from setuptools import setup, find_packages
 
 setup(
     name='hq_superset',
-    version='0.2.0',
-    description='CommCareHQ Superset Integration',
+    version='0.3.0',
+    description='CommCare HQ Superset Integration',
     license='Apache2',
-    author='Dimagi Inc',
+    author='Dimagi Inc.',
     author_email='sreddy@dimagi.com',
     url='https://github.com/dimagi/hq_superset',
     packages=find_packages(exclude=['docs', 'tests']),
     include_package_data=True,
     install_requires=[
-        # Werkzeug 2.1 doesn't work, so pin it
-        'Werkzeug==2.0.0',
-        'jinja2==3.0.3',
-        'dimagi-superset==2.0.1',
-        'authlib==1.0.1',
-        'requests==2.28.1',
-        'psycopg2==2.9.3',
-        'WTForms==2.3.3',
-        'cryptography==37.0.4',
-        'sentry-sdk==1.9.10',
+        'apache-superset==3.1.0',
+        # Dependencies based on Superset 3.1.0 where applicable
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -1,35 +1,32 @@
...
-        'sentry-sdk==1.9.10',
+        'apache-superset==3.1.0',
+        # Dependencies based on Superset 3.1.0 where applicable
+        'Authlib==1.3.0',
         'celery==5.2.7',
+        'cryptography==41.0.2',
+        'jinja2==3.1.2',
+        'psycopg2==2.9.6',
+        'requests==2.31.0',
+        'sentry-sdk==1.39.2',
```
SmittieC, line 22: If we use `sentry-sdk[flask]`, then we don't have to install `blinker`

SmittieC (approved):
This is gold, thanks @kaapstorm

Charl1996:
Nice!

setup.py
kaapstorm: 1f4763b

setup.py
```diff
@@ -1,35 +1,32 @@
...
-        'Werkzeug==2.0.0',
-        'jinja2==3.0.3',
-        'dimagi-superset==2.0.1',
-        'authlib==1.0.1',
-        'requests==2.28.1',
-        'psycopg2==2.9.3',
-        'WTForms==2.3.3',
-        'cryptography==37.0.4',
-        'sentry-sdk==1.9.10',
+        'apache-superset==3.1.0',
```
sravfeyn, line 14: Shouldn't we be using our forked superset? This would break some of the features we have implemented already. Specifically this https://github.com/dimagi/superset/pull/1

sravfeyn: There is few docs I added to keep our fork updated. https://github.com/dimagi/hq_superset/blob/master/apache-superset.md
We would need to use the fork repo instead of the core repo.

kaapstorm: Thank you for the feedback!

This resolves some of my confusion: As you can see from the [master branch of dimagi-superset](https://github.com/dimagi/superset), there is [only one commit difference between it and apache-superset](https://github.com/apache/superset/compare/master...dimagi:superset:master). And that commit has some Portuguese translations (which cannot be rebased on top of Superset 3.1.0 because of unresolveable conflicts).

It introduces more questions though. I'll follow them up offline. I'm very sorry that I missed that documentation!

mkangia: hey @kaapstorm 
Just confirming that we are now using the apache superset only or its a temporary change?

kaapstorm: It's a temporary change. I have uploaded dimagi-superset 3.1.0 to PyPI just now, and I'll update these requirements.

kaapstorm: https://github.com/dimagi/hq_superset/pull/36

PR #38, opened 2024-02-07
Update `apache-superset.md` docs

kaapstorm:
Updates documentation with recent lessons learned.

(Documentation-only changes.)

### DIFF
--- a/apache-superset.md
+++ b/apache-superset.md
@@ -1,126 +1,270 @@
-## Dimagi Superset Fork
+Dimagi Superset Fork
+--------------------
 
-This is a fork of https://github.com/apache/superset. This repo is
-maintained to build a slightly customized version of the original
-[Superset PyPI package](https://pypi.org/project/apache-superset/).
+[Dimagi Superset](https://github.com/dimagi/hq_superset) is a fork of
+[Apache Superset](https://github.com/apache/superset). Dimagi Superset
+is maintained to build a slightly customized version of the
+[Apache Superset PyPI package](https://pypi.org/project/apache-superset/).
 
 
 ### Making changes
 
-We build most of the customization and integration inside
-https://github.com/dimagi/hq_superset and avoid customizing this repo
-directly so that we don't diverge too much from original Superset. But
-we do that when it's very complicated to do it inside hq_superset and
-when it's super easy to do it in this fork.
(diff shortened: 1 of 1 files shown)


apache-superset.md
```diff
@@ -1,86 +1,285 @@
...
+`dimagi_pt` branch of
+[Dimagi Superset] (https://github.com/dimagi/superset).
+
+If you get a new version of Dimagi's Portuguese translations,
+please _replace_ the last `dimagi_pt_messages.po` file, and commit it
+in the `dimagi_pt` branch.
+
+To generate updated language files:
+
+0. (Optional) Set up a virtual Node environment (like a Python virtual
```
mkangia, line 139: just curious about the 0 here.

apache-superset.md
```diff
@@ -1,86 +1,285 @@
...
+
+- **msgmerge**, which is probably already be installed. It is included
+  in the **gettext** package, which is one of the pre-install
+  requirements for CommCare HQ dev environments on Linux.
+
+- **po2json**, which needs to be installed using **npm**.
+
+You will also need Dimagi's Portuguese translations. You can reach out
+to members of the Solutions Division who work on projects in Mozambique
+for an updated version. The current version should be named
```
mkangia, line 128: If possible, good to take a note of where/how this translation is done, may be on confluence and link that here.
Assuming if someone needs to do this in future, they would be confused on where to look next after reading this.

mkangia (approved):

SmittieC (approved):
Thanks for this!

apache-superset.md
kaapstorm: I used a "0" to emphasize that this is a step before we begin. (A google will show that "Step 0" or "Step Zero" is generally understood to mean a preliminary step.)

I use nodeenv because I've encountered Node library version conflicts and not known what needs which version. It avoids those problems. But it doesn't seem to be used in the Node community. It's a Python tool, and I'm not sure it's even widely used in the Python community. Definitely optional (regardless of how much I appreciate it :) ).

apache-superset.md
kaapstorm: Great idea! I don't think that exists yet, but I'll reach out, and see if they can create that Confluence page.

PR #39, opened 2024-02-14
Permission for domains to edit virtual datasets as well

SmittieC:
A number of Analytics users have been telling us they can create virtual datasets, but not edit them. This PR's work is based off of  [this PR](https://github.com/dimagi/hq_superset/pull/26) that Zandre did last year, but at that time it was deemed unnecessary.

This change isn't critical, so it can wait until the base branch is merged (nh/webhook). The reason I branched off of this branch is because there's some config issues with the tests, so I don't have to first fix those and cause potential conflicts.


hq_superset/utils.py
```diff
@@ -182,13 +182,24 @@ def re_eval_roles(self, existing_roles, new_domain_role):
...
     def sync_domain_role(self, domain):
         # This creates DB schema, role and schema permissions for the domain and
         #   assigns the role to the current_user
         self._ensure_schema_created(domain)
         permission = self._ensure_schema_perm_created(domain)
         role = self._ensure_domain_role_created(domain)
         self.sm.add_permission_role(role, permission)
+
+        ds_permission = self._ensure_datasource_perm_created()
+        self.sm.add_permission_role(role, ds_permission)
```
mkangia, line 201: Just checking if we should put this in the role created for the domain?
I would prefer if that role is only limited to allowing access to domain schema.

We could simply add a new role for this and add that to the user.

Though, i think its better to solve this via https://dimagi.atlassian.net/browse/SC-2853 so additional permissions are handled the same way. We could add this permission to the permissions listed [here](https://docs.google.com/document/d/1AybtXCthYwmq3tLl8ygD3tX1hD79wj9JKoTOtzpfoW4/edit#heading=h.jnpvevto1u08)

What do you think?

SmittieC: Yes! I forgot that this was a thing. I agree, let's handle this through that feature

SmittieC:
Closing this PR in favor of handling this through [this feature](https://dimagi.atlassian.net/browse/SC-2853)

hq_superset/utils.py
mkangia: thanks for the flexibility @SmittieC

PR #40, opened 2024-02-14
Proposal: Add ruff (linter and formatter)

SmittieC:
Things will get [ruff](https://github.com/astral-sh/ruff) if we don't add something like this sooner or later. This is a linter and formatter. This is apparently way faster than tools like `black` and `flake8`.

I also added a Makefile to reduce the amount of commands one has to run each time working with this.

It's base branch is my latest branch based off of `nh/webhook`. This is to avoid as much merge conflicts as possible if/when this is merged


hq_superset/__init__.py
```diff
@@ -10,36 +10,36 @@ def flask_app_mutator(app):
...
     # Allow loading templates from the templates directory in this project as well
 
-    template_path = os.sep.join((
-        os.path.dirname(os.path.abspath(__file__)),
-        'templates'
-    ))
-    my_loader = jinja2.ChoiceLoader([
+    template_path = os.sep.join(
+        (os.path.dirname(os.path.abspath(__file__)), "templates")
+    )
```
kaapstorm, line 32: No.

hq_superset/api.py
```diff
@@ -36,8 +36,8 @@ def save_token(token, request):
     tok = Token(
         client_id=client.client_id,
         expires_at=expires_at,
-        access_token=token['access_token'],
-        token_type=token['token_type'],
+        access_token=token["access_token"],
```
kaapstorm, line 39: Absolutely not.

kaapstorm (changes requested):
I love this idea. I am a fan of ruff. I use it in my own projects. It's the best linter.

But I will push back hard on a couple of those formatting choices: Either use single quotes, or leave the quotes alone. Also, these bracket decisions do not improve readability.

hq_superset/__init__.py
SmittieC: Haha agreed!

pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[tool.ruff]
+line-length = 120
```
SmittieC, line 2: I can't remember, do we as dimagi use 120 chars as the line length?

hq_superset/api.py
SmittieC: You might find https://github.com/dimagi/hq_superset/pull/40/commits/a8404412fdaf309d3b20afa0efa99b78978be89c to your liking?

pyproject.toml
zandre-eng: For flake8, it's currently set to 115 lines for CommCare HQ ([code reference](https://github.com/dimagi/commcare-hq/blob/4274186f9cb6b3e337fe104c8e2019ec13a0f29d/.flake8#L2))

kaapstorm: We use 115 (to make code review easier (if your browser is maximized on a 1280-wide screen)).

(Personally, I prefer PEP-8 line lengths (79 for code, 72 for docstrings) because (1) it makes code readable when you split your windows vertically -- code on the left, tests on the right -- and (2) long lines of text are ugly and hard to read, which is why newspapers have columns.)

pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[tool.ruff]
+line-length = 115
```
kaapstorm, line 2: Curious what you think of this:

```suggestion
line-length = 79  # Preferred line length to wrap lines

[tool.ruff.lint.pycodestyle]
max-line-length = 115  # Line length to allow for line-too-long violations
```

I noticed that Ruff will not reformat strings if they exceed `line-length`.

SmittieC: 79 is probably fine.

> I noticed that Ruff will not reformat strings if they exceed line-length.

Weird

pyproject.toml
```diff
@@ -0,0 +1,5 @@
...
+
+[tool.ruff.format]
+quote-style = "single"
```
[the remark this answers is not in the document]
Charl1996: I find it ironic that `"single"` is in double quotes.

SmittieC:
> On principle I think this is a good idea. I think this is also a good codebase to use this in. And I'm coming around to liking the idea of having only one choice of quotation marks. I prefer single, but I could also make an argument in favor of double. ... I do like 79-character lines though, because it makes my IDE layout happy.

Cool. I don't have any objections to 79 character lines. I'll make an update

pyproject.toml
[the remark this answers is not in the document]
SmittieC: Yeah that's my reasoning as well. I don't mind single or double quotes. I presume you prefer single quotes?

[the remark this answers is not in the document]
SmittieC: Oh I see you said it just below: "I prefer single"

mkangia:
Do we still need isort on top of this? I like that it fixes files automatically

Does ruffle also take care of line spacing between methods? I have always used 2 spaces and now assume that is the standard.

mkangia:
PR is still in draft though.

SmittieC:
FYI: I snuck https://github.com/dimagi/hq_superset/pull/40/commits/d1ee5ae363fb56070c3a208a26a3252ab214b8bf into this PR.

SmittieC:
> Do we still need isort on top of this? I like that it fixes files automatically
> 
> Does ruffle also take care of line spacing between methods? I have always used 2 spaces and now assume that is the standard.

Looks like we can get sorting with https://github.com/dimagi/hq_superset/pull/40/commits/8aec7d95f2fb45c53596534c4949a5056cf81926. As for the spaces between methods, I'm not 100% sure. Will have to look into this a bit more, but I'm sure one can achieve this.

mkangia:
> Looks like we can get sorting with https://github.com/dimagi/hq_superset/commit/8aec7d95f2fb45c53596534c4949a5056cf81926.

Perfect

Charl1996:
@SmittieC What is the status of this PR?

kaapstorm (approved):
Happy

PR #41, opened 2024-02-27
Background sync for datasets from CommCare HQ

kaapstorm:
Jira: [SC-3212](https://dimagi.atlassian.net/browse/SC-3212)

This pull request includes
* A series of small prefactor commits
* A large commit that adds an API endpoint for accepting dataset change requests from CommCare HQ
* A large commit that adds an OAuth 2.0 server for authenticating incoming requests
* A migration for adding the models for the OAuth 2.0 server

Easiest to review :blowfish: commit :octopus: by :tropical_fish: commit :fish: 

This code has been tested locally.

It will be deployed to Staging on Wed Feb 28.

QA is scheduled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,10 @@ admin user, for example, to add a database connection. To enable local
 user authentication, in `superset_config.py`, set
 `AUTH_TYPE = AUTH_DB`.
 
+Doing this will prevent CommCare HQ users from logging in, so it should
+only be done in production environments when CommCare Analytics is not
+in use.
+
 To return to allowing CommCare HQ users to log in, set it back to
 `AUTH_TYPE = AUTH_OAUTH`.
 
--- a/README.md
+++ b/README.md
@@ -102,8 +102,19 @@ server. (Yes, it's "OAUTHLIB" this time, not "AUTHLIB" as before.)
 $ export OAUTHLIB_INSECURE_TRANSPORT=1
 ```
 
-### Importing UCRs using Redis and Celery
 
+### Logging in as a local admin user
+
+There might be situations where you need to log into Superset as a local
+admin user, for example, to add a database connection. To enable local
+user authentication, in `superset_config.py`, set
+`AUTH_TYPE = AUTH_DB`.
+
+To return to allowing CommCare HQ users to log in, set it back to
+`AUTH_TYPE = AUTH_OAUTH`.
+
+
+### Importing UCRs using Redis and Celery
 
 Celery is used to import UCRs that are larger than
 `hq_superset.views.ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES`. If you need
--- a/README.md
+++ b/README.md
@@ -6,9 +6,14 @@ This is a Python package that integrates Superset and CommCare HQ.
 Local Development
 -----------------
 
-Follow below instructions.
+### Preparing CommCare HQ
 
-### Setup env
+The 'User configurable reports UI' feature flag must be enabled for the
+domain in CommCare HQ, even if the data sources to be imported were
+created by Report Builder, not a UCR.
+
+
+### Setting up a dev environment
 
 While doing development on top of this integration, it's useful to
 install this via `pip -e` option so that any changes made get reflected
@@ -51,11 +56,12 @@ directly without another `pip install`.
 Read through the initialization instructions at
 https://superset.apache.org/docs/installation/installing-superset-from-scratch/#installing-and-initializing-superset.
 
-Create the database. These instructions assume that PostgreSQL is
-running on localhost, and that its user is "commcarehq". Adapt
-accordingly:
+Create a database for Superset, and a database for storing data from
+CommCare HQ. Adapt the username and database names to suit your
+environment.
 ```bash
-$ createdb -h localhost -p 5432 -U commcarehq superset_meta
+$ createdb -h localhost -p 5432 -U postgres superset
+$ createdb -h localhost -p 5432 -U postgres superset_hq_data
 ```
 
 Set the following environment variables:
@@ -64,10 +70,17 @@ $ export FLASK_APP=superset
 $ export SUPERSET_CONFIG_PATH=/path/to/superset_config.py
 ```
 
-Initialize the database. Create an administrator. Create default roles
+Set this environment variable to allow OAuth 2.0 authentication with
+CommCare HQ over insecure HTTP. (DO NOT USE THIS IN PRODUCTION.)
+```bash
+$ export AUTHLIB_INSECURE_TRANSPORT=1
+```
+
+Initialize the databases. Create an administrator. Create default roles
 and permissions:
 ```bash
 $ superset db upgrade
+$ superset db upgrade --directory hq_superset/migrations/
 $ superset fab create-admin
 $ superset load_examples  # (Optional)
 $ superset init
@@ -78,28 +91,16 @@ You should now be able to run superset using the `superset run` command:
 ```bash
 $ superset run -p 8088 --with-threads --reload --debugger
 ```
-However, OAuth login does not work yet as hq-superset needs a Postgres
-database created to store CommCare HQ data.
 
+You can now log in as a CommCare HQ web user.
 
-### Create a Postgres Database Connection for storing HQ data
-
-- Create a Postgres database. e.g.
-  ```bash
-  $ createdb -h localhost -p 5432 -U commcarehq hq_data
-  ```
-- Log into Superset as the admin user created in the Superset
-  installation and initialization. Note that you will need to update
-  `AUTH_TYPE = AUTH_DB` to log in as admin user. `AUTH_TYPE` should be
-  otherwise set to `AUTH_OAUTH`.
-- Go to 'Data' -> 'Databases' or http://127.0.0.1:8088/databaseview/list/
-- Create a database connection by clicking '+ DATABASE' button at the top.
-- The name of the DISPLAY NAME should be 'HQ Data' exactly, as this is
-  the name by which this codebase refers to the Postgres DB.
-
-OAuth integration should now be working. You can log in as a CommCare
-HQ web user.
-
+In order for CommCare HQ to sync data source changes, you will need to
+allow OAuth 2.0 authentication over insecure HTTP. (DO NOT USE THIS IN
+PRODUCTION.) Set this environment variable in your CommCare HQ Django
+server. (Yes, it's "OAUTHLIB" this time, not "AUTHLIB" as before.)
+```bash
+$ export OAUTHLIB_INSECURE_TRANSPORT=1
+```
 
 ### Importing UCRs using Redis and Celery
 
@@ -127,6 +128,41 @@ Tests use pytest, which is included in `requirements_dev.txt`:
 The test runner can only run tests that do not import from Superset. The
 code you want to test will need to be in a module whose dependencies
 don't include Superset.
+
+
+### Creating a migration
+
+You will need to create an Alembic migration for any new SQLAlchemy
+models that you add. The Superset CLI should allow you to do this:
+
+```shell
+$ superset db revision --autogenerate -m "Add table for Foo model"
+```
+
+However, problems with this approach have occurred in the past. You
+might have more success by using Alembic directly. You will need to
+modify the configuration a little to do this:
+
+1. Copy the "HQ_DATA" database URI from `superset_config.py`.
+
+2. Paste it as the value of `sqlalchemy.url` in
+   `hq_superset/migrations/alembic.ini`.
+
+3. Edit `env.py` and comment out the following lines:
+   ```
+   hq_data_uri = current_app.config['SQLALCHEMY_BINDS'][HQ_DATA]
+   decoded_uri = urllib.parse.unquote(hq_data_uri)
+   config.set_main_option('sqlalchemy.url', decoded_uri)
+   ```
+
+Those changes will allow Alembic to connect to the "HD Data" database
+without the need to instantiate Superset's Flask app. You can now
+autogenerate your new table with:
+
+```shell
+$ cd hq_superset/migrations/
+$ alembic revision --autogenerate -m "Add table for Foo model"
+```
 
 
 Upgrading Superset
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -9,10 +9,11 @@ def flask_app_mutator(app):
     # return
     from superset.extensions import appbuilder
 
-    from . import hq_domain, views
+    from . import api, hq_domain, views
 
     appbuilder.add_view(views.HQDatasourceView, 'Update HQ Datasource', menu_cond=lambda *_: False)
     appbuilder.add_view(views.SelectDomainView, 'Select a Domain', menu_cond=lambda *_: False)
+    appbuilder.add_api(api.DataSetChangeAPI)
     app.before_request_funcs.setdefault(None, []).append(
         hq_domain.before_request_hook
     )
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -17,12 +17,8 @@ def flask_app_mutator(app):
     appbuilder.add_api(api.DataSetChangeAPI)
     oauth2_server.config_oauth2(app)
 
-    app.before_request_funcs.setdefault(None, []).append(
-        hq_domain.before_request_hook
-    )
-    app.after_request_funcs.setdefault(None, []).append(
-        hq_domain.after_request_hook
-    )
+    app.before_request(hq_domain.before_request_hook)
+    app.after_request(hq_domain.after_request_hook)
     app.strict_slashes = False
     override_jinja2_template_loader(app)
 
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -9,11 +9,14 @@ def flask_app_mutator(app):
     # return
     from superset.extensions import appbuilder
 
-    from . import api, hq_domain, views
+    from . import api, hq_domain, oauth2_server, views
 
     appbuilder.add_view(views.HQDatasourceView, 'Update HQ Datasource', menu_cond=lambda *_: False)
     appbuilder.add_view(views.SelectDomainView, 'Select a Domain', menu_cond=lambda *_: False)
+    appbuilder.add_api(api.OAuth)
     appbuilder.add_api(api.DataSetChangeAPI)
+    oauth2_server.config_oauth2(app)
+
     app.before_request_funcs.setdefault(None, []).append(
         hq_domain.before_request_hook
     )
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -22,6 +22,13 @@ def flask_app_mutator(app):
     app.strict_slashes = False
     override_jinja2_template_loader(app)
 
+    # A proxy (maybe) is changing the URL scheme from "https" to "http"
+    # on commcare-analytics-staging.dimagi.com, which breaks the OAuth
+    # 2.0 secure transport check despite transport being over HTTPS. I
+    # hate to do this, but werkzeug.contrib.fixers.ProxyFix didn't fix
+    # it. So I've run out of better options. (Norman 2024-03-13)
+    os.environ['AUTHLIB_INSECURE_TRANSPORT'] = '1'
+
 
 def override_jinja2_template_loader(app):
     # Allow loading templates from the templates directory in this project as well
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -1,9 +1,9 @@
 import json
 from http import HTTPStatus
 
-from flask import request
-from flask_appbuilder.api import BaseApi
-from flask_appbuilder.baseviews import expose
+from flask import jsonify, request
+from flask_appbuilder.api import BaseApi, expose
+from sqlalchemy.orm.exc import NoResultFound
 from superset.superset_typing import FlaskResponse
 from superset.views.base import (
     handle_api_exception,
@@ -12,6 +12,27 @@ from superset.views.base import (
 )
 
 from .models import DataSetChange
+from .oauth2_server import authorization, require_oauth
+
+
+class OAuth(BaseApi):
+
+    def __init__(self):
+        super().__init__()
+        self.route_base = "/oauth"
+
+    @expose("/token", methods=('POST',))
+    def issue_access_token(self):
+        try:
+            response = authorization.create_token_response()
+        except NoResultFound:
+            return jsonify({"error": "Invalid client"}), 401
+
+        if response.status_code >= 400:
+            return response
+
+        data = json.loads(response.data.decode("utf-8"))
+        return jsonify(data)
 
 
 class DataSetChangeAPI(BaseApi):
@@ -28,6 +49,7 @@ class DataSetChangeAPI(BaseApi):
 
     @expose('/change/', methods=('POST',))
     @handle_api_exception
+    @require_oauth()
     def post_dataset_change(self) -> FlaskResponse:
         if request.content_length > self.MAX_REQUEST_LENGTH:
             return json_error_response(
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -1,0 +1,47 @@
+import json
+from http import HTTPStatus
+
+from flask import request
+from flask_appbuilder.api import BaseApi
+from flask_appbuilder.baseviews import expose
+from superset.superset_typing import FlaskResponse
+from superset.views.base import (
+    handle_api_exception,
+    json_error_response,
+    json_success,
+)
+
+from .models import DataSetChange
+
+
+class DataSetChangeAPI(BaseApi):
+    """
+    Accepts changes to datasets from CommCare HQ data forwarding
+    """
+
+    MAX_REQUEST_LENGTH = 10 * 1024 * 1024  # reject JSON requests > 10MB
+
+    def __init__(self):
+        self.route_base = '/hq_webhook'
+        self.default_view = 'post_dataset_change'
+        super().__init__()
+
+    @expose('/change/', methods=('POST',))
+    @handle_api_exception
+    def post_dataset_change(self) -> FlaskResponse:
+        if request.content_length > self.MAX_REQUEST_LENGTH:
+            return json_error_response(
+                HTTPStatus.REQUEST_ENTITY_TOO_LARGE.description,
+                status=HTTPStatus.REQUEST_ENTITY_TOO_LARGE.value,
+            )
+
+        try:
+            request_json = json.loads(request.get_data(as_text=True))
+            change = DataSetChange(**request_json)
+            change.update_dataset()
+            return json_success('Dataset updated')
+        except json.JSONDecodeError:
+            return json_error_response(
+                'Invalid JSON syntax',
+                status=HTTPStatus.BAD_REQUEST.value,
+            )
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -43,7 +43,7 @@ class DataSetChangeAPI(BaseApi):
     MAX_REQUEST_LENGTH = 10 * 1024 * 1024  # reject JSON requests > 10MB
 
     def __init__(self):
-        self.route_base = '/hq_webhook'
+        self.route_base = '/commcarehq_dataset'
         self.default_view = 'post_dataset_change'
         super().__init__()
 
--- a/hq_superset/const.py
+++ b/hq_superset/const.py
@@ -1,0 +1,4 @@
+# The name of the database for storing data related to CommCare HQ
+HQ_DATABASE_NAME = "HQ Data"
+
+OAUTH2_DATABASE_NAME = "oauth2-server-data"
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -26,6 +26,7 @@ DOMAIN_EXCLUDED_VIEWS = [
     "AuthDBView.login",
     "SelectDomainView.list",
     "SelectDomainView.select",
+    "DataSetChangeAPI.post_dataset_change",
     "appbuilder.static",
     "static",
 ]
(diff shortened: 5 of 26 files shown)


hq_superset/models.py
```diff
@@ -0,0 +1,34 @@
...
+            database.get_sqla_engine_with_context() as engine,
+            engine.connect() as connection,
+            connection.begin()  # Commit on leaving context
+        ):
+            delete_stmt = table.delete().where(table.c.doc_id == self.doc_id)
+            connection.execute(delete_stmt)
+            if self.data:
+                rows = list(cast_data_for_table(self.data, table))
+                insert_stmt = table.insert().values(rows)
+                connection.execute(insert_stmt)
```
mkangia, line 47: is this deleting and then inserting the row for the same case instead of an update?
Not that I am opposed to the approach, just trying to understand the reasoning behind it.

hq_superset/api.py
```diff
@@ -0,0 +1,47 @@
...
+
+class DataSetChangeAPI(BaseApi):
+    """
+    Accepts changes to datasets from CommCare HQ data forwarding
+    """
+
+    MAX_REQUEST_LENGTH = 10 * 1024 * 1024  # reject JSON requests > 10MB
+
+    def __init__(self):
+        self.route_base = '/hq_webhook'
```
mkangia, line 46: nit: do we want to keep a better url for this? I assume this is getting used internally and I don't fully understand where all this will be used. "webhook" seems quite open ended on what that could mean.

hq_superset/const.py
```diff
@@ -0,0 +1,2 @@
+# The name of the database for storing data related to CommCare HQ
+HQ_DATA = "HQ Data"
```
mkangia, line 2: nit: `CommCareHQ Data`?

mkangia (commented):
thanks for the refactor commits @kaapstorm.

I tried reviewing this PR but it was a bit difficult with this PR attending to quite a lot of changes with some big commits.
Plus the part about OAuth that came along with syncing data sources is new to me.

Are you able to break this down further for easier review? Or else if its better the way it is, I will take a more dedicated pass. Please let me know

hq_superset/const.py
kaapstorm: Reply nit: "CommCare HQ" [has a space in it](https://www.commcarehq.org/). :joy: 

Unfortunately this value needs to match the database name in existing environments. Previously, users had to create this database manually (see [instructions in README.md](https://github.com/dimagi/hq_superset/blob/master/README.md#create-a-postgres-database-connection-for-storing-hq-data), specifically, "The name of the DISPLAY NAME should be 'HQ Data' exactly". So we're stuck with "HQ Data" (even though the convention in Superset is to use a lowercase word, like "commcare-hq-data").

hq_superset/models.py
kaapstorm: I wrangled with this for a long time. SQLAlchemy's [Postgres dialect supports upsert](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#insert-on-conflict-upsert). But there is a catch: The "doc_id" column must be indexed in order to use it as a constraint in the upsert's "where" clause.

That table is created by Pandas, and Pandas can't create an index on that column. You can create the table first, and then dump Pandas data into it, but I couldn't get Pandas to create the rest of columns and populate the table correctly. Maybe I was missing something. Eventually I decided to just delete and then insert.

Next time I have a few more hours to spend on it, I'll go back to it. Or we can make a follow-up ticket for nice-to-haves.

hq_superset/api.py
kaapstorm: Good point. "hq_datasource"? "hq_dataset"?

mkangia: how about just `dataset`? I see we have been following a convention to put a `hq_` in all names.
Is it to avoid conflict with superset routes?

mkangia: having `hq_` in all names makes me think that its coming from hq but hq does not have anything called dataset.
hence you also suggested `hq_datasource` but this view is changing dataset.

kaapstorm:
Tests are failing because they are not configured correctly. I'll fix that.

@mkangia you asked:
> Are you able to break this down further for easier review? 

I agree that [Add OAuth 2.0 server](https://github.com/dimagi/hq_superset/pull/41/commits/dad8b91badb6c1d10d1f564f7bc585154a7ee842) is a monster. All the commits that went into it [can be found here](https://github.com/dimagi/hq_superset/pull/29) but I'm not sure that's any more helpful.

I'll try to unpack them. I'll timebox it and see how I get on. (Rearranging and squashing the commits from the draft PR took hours. Hopefully unpacking might be quicker.)

mkangia:
Thanks @kaapstorm 

Good to timebox or else then If its easier, we could jump on a call where we can walk through the big commits

hq_superset/api.py
kaapstorm: > Is it to avoid conflict with superset routes?

Yes, exactly. Superset already has "dataset" routes.

> but this view is changing dataset.

Yes. ... Although it is the endpoint for DataSourceRepeaters. The URL for their ConnectionSettings will be 'hq_datasource/change". So I guess at least it makes sense in the context of the client/HQ.

mkangia: > Although it is the endpoint for DataSourceRepeaters

I am thinking that we should name things in here independent of what they are used for in CommCareHQ and more around what they mean here. I guess I will have a better suggestion of what this should be named if I look into this more deeply. Fine to leave it as it is then.
We are in the early phases of this product so would be nice to focus on good names early on.

hq_superset/api.py
```diff
@@ -0,0 +1,47 @@
...
+
+from .models import DataSetChange
+
+
+class DataSetChangeAPI(BaseApi):
+    """
+    Accepts changes to datasets from CommCare HQ data forwarding
+    """
+
+    MAX_REQUEST_LENGTH = 10 * 1024 * 1024  # reject JSON requests > 10MB
```
Charl1996, line 43: @kaapstorm How was this limit determined?

hq_superset/api.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(self):
+        self.route_base = '/hq_webhook'
+        self.default_view = 'post_dataset_change'
+        super().__init__()
+
+    @expose('/change/', methods=('POST',))
+    @handle_api_exception
+    def post_dataset_change(self) -> FlaskResponse:
```
Charl1996, line 53: @kaapstorm Quickly remind me again, is this endpoint going to be hit for every row change in a data source?

hq_superset/utils.py
```diff
@@ -193,3 +204,52 @@ def array_is_falsy(array_values):
         return []
 
     return array_values
+
+
+def js_to_py_datetime(jsdt, preserve_tz=True):
```
Charl1996, line 214: The docstring is amazing!

hq_superset/models.py
```diff
@@ -32,3 +45,71 @@ def update_dataset(self):
                 rows = list(cast_data_for_table(self.data, table))
                 insert_stmt = table.insert().values(rows)
                 connection.execute(insert_stmt)
+
+
+class HQClient(db.Model, OAuth2ClientMixin):
```
Charl1996, line 50: I'm going to be nitpicking here...

Can we maybe call this something like `HQOAuthClient`? The reason is that the current naming made me think this was the request client (don't know if I'm just super pedantic).

Anyway, I'll leave it to your own discretion...

hq_superset/services.py
```diff
@@ -144,6 +149,35 @@ def to_sql(df, replace=False):
         raise ex
 
 
+def subscribe_to_hq_datasource(domain, datasource_id):
+    hq_client = HQClient.get_by_domain(domain)
+    if hq_client is None:
+        hq_client = HQClient.create_domain_client(domain)
```
Charl1996, line 155: Nit: can this be a separate function? (maybe something like `get_oauth_client_for_domain`)

hq_superset/services.py
```diff
@@ -144,6 +149,35 @@ def to_sql(df, replace=False):
...
+
+    hq_request = HQRequest(url=datasource_subscribe(domain, datasource_id))
+    webhook_url = urljoin(
+        request.root_url,
+        url_for('DataSetChangeAPI.post_dataset_change'),
+    )
```
Charl1996, line 161: Personally I'd put this in another function as well, just so `subscribe_to_hq_datasource` doens't have to know how the `webhook_url` is generated., but I'll leave it up to you...

README.md
```diff
@@ -6,9 +6,14 @@ This is a Python package that integrates Superset and CommCare HQ.
 Local Development
 -----------------
 
-Follow below instructions.
+### Preparing CommCare HQ
 
-### Setup env
+The 'User configurable reports UI' feature flag must be enabled for the
```
Charl1996, line 11: Maybe more of an HQ discussion, but should we not have a different FF for this workflow so users don't have to be concerned about the `User Configurable Reports UI` feature flag until they actually want to user the UCRs?

Charl1996:
Absolutely great work, @kaapstorm !

Just a couple of comments, nothing big.

hq_superset/models.py
mkangia: thanks @kaapstorm 
the current approach sounds good to me

kaapstorm:
Please excuse the force-pushes. Ready for review @mkangia .

And thanks for the review @Charl1996 ! I'll follow up in the morning.

hq_superset/models.py
```diff
@@ -32,3 +45,71 @@ def update_dataset(self):
                 rows = list(cast_data_for_table(self.data, table))
                 insert_stmt = table.insert().values(rows)
                 connection.execute(insert_stmt)
+
+
+class OAuth2Client(db.Model, OAuth2ClientMixin):
+    __bind_key__ = OAUTH2_DATABASE_NAME
+    __tablename__ = 'hq_oauth_client'
```
mkangia, line 52: nit: building on the conversation about naming, this could be `commcarehq_oauth_client` or `commcare_hq_oauth_client`

hq_superset/models.py
```diff
@@ -32,3 +45,71 @@ def update_dataset(self):
...
+    def check_client_secret(self, plaintext):
+        return self.get_client_secret() == plaintext
+
+    def revoke_tokens(self):
+        revoked_at = int(time.time())
+        stmt = (
+            update(OAuth2Token)
+            .where(OAuth2Token.client_id == self.client_id)
+            .where(OAuth2Token.access_token_revoked_at == 0)
+            .values(access_token_revoked_at=revoked_at)
```
mkangia, line 82: Can we add a docstring to this function please?
I assume it simply revokes all current active tokens for the client.

And since I don't understand this syntax well, just confirming that we don't need `values(OAuth2Token.access_token_revoked_at=revoked_at)`?

hq_superset/services.py
```diff
@@ -145,6 +151,35 @@ def dataframe_to_sql(df, replace=False):
...
 
 
+def subscribe_to_hq_datasource(domain, datasource_id):
+    hq_client = OAuth2Client.get_by_domain(domain)
+    if hq_client is None:
+        hq_client = OAuth2Client.create_domain_client(domain)
+
+    hq_request = HQRequest(url=datasource_subscribe(domain, datasource_id))
+    webhook_url = urljoin(
+        request.root_url,
```
mkangia, line 161: nit: should this be `hq_request.root_url`?
and same later in this function?

README.md
```diff
@@ -64,10 +70,17 @@ $ export FLASK_APP=superset
 $ export SUPERSET_CONFIG_PATH=/path/to/superset_config.py
 ```
 
-Initialize the database. Create an administrator. Create default roles
+Set this environment variable to allow OAuth 2.0 authentication with
+CommCare HQ over insecure HTTP. (DO NOT USE THIS IN PRODUCTION.)
+```bash
+$ export AUTHLIB_INSECURE_TRANSPORT=1
```
mkangia, line 76: Is this and `OAUTHLIB_INSECURE_TRANSPORT` needed if we are using OAuth with CommCareHQ environment available locally?
If we are providing this option to communicate with live CommCareHQ environments, we probably should not.

README.md
```diff
@@ -129,6 +130,41 @@ code you want to test will need to be in a module whose dependencies
...
+   config.set_main_option('sqlalchemy.url', decoded_uri)
+   ```
+
+Those changes will allow Alembic to connect to the "HD Data" database
+without the need to instantiate Superset's Flask app. You can now
+autogenerate your new table with:
+
+```shell
+$ cd hq_superset/migrations/
+$ alembic revision --autogenerate -m "Add table for Foo model"
```
mkangia, line 164: Nice!

hq_superset/migrations/env.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+        with context.begin_transaction():
+            context.run_migrations()
+
+
+if context.is_offline_mode():
+    run_migrations_offline()
+else:
+    run_migrations_online()
```
mkangia, line 77: Did we take this process from somewhere?
Just thinking why we would need offline/online processes.

README.md
```diff
@@ -129,6 +130,41 @@ code you want to test will need to be in a module whose dependencies
...
+
+3. Edit `env.py` and comment out the following lines:
+   ```
+   hq_data_uri = current_app.config['SQLALCHEMY_BINDS'][HQ_DATA]
+   decoded_uri = urllib.parse.unquote(hq_data_uri)
+   config.set_main_option('sqlalchemy.url', decoded_uri)
+   ```
+
+Those changes will allow Alembic to connect to the "HD Data" database
+without the need to instantiate Superset's Flask app. You can now
```
mkangia, line 159: and then one should uncomment them back afterwards?

README.md
```diff
@@ -51,11 +56,12 @@ directly without another `pip install`.
...
 
-Create the database. These instructions assume that PostgreSQL is
-running on localhost, and that its user is "commcarehq". Adapt
-accordingly:
+Create a database for Superset, and a database for storing data from
+CommCare HQ. Adapt the username and database names to suit your
+environment.
 ```bash
-$ createdb -h localhost -p 5432 -U commcarehq superset_meta
+$ createdb -h localhost -p 5432 -U postgres superset
```
mkangia, line 63: did we rename `superset_meta` to `superset`?

README.md
```diff
@@ -78,28 +91,16 @@ You should now be able to run superset using the `superset run` command:
...
-### Create a Postgres Database Connection for storing HQ data
-
-- Create a Postgres database. e.g.
-  ```bash
-  $ createdb -h localhost -p 5432 -U commcarehq hq_data
-  ```
-- Log into Superset as the admin user created in the Superset
-  installation and initialization. Note that you will need to update
-  `AUTH_TYPE = AUTH_DB` to log in as admin user. `AUTH_TYPE` should be
-  otherwise set to `AUTH_OAUTH`.
```
mkangia, line 94: Just checking on why we removed this part?

hq_superset/migrations/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
mkangia, line 1: nit: we need this file?

hq_superset/migrations/alembic.ini
```diff
@@ -0,0 +1,115 @@
...
+# set to 'true' to search source files recursively
+# in each "version_locations" directory
+# new in Alembic version 1.10
+# recursive_version_locations = false
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+sqlalchemy.url = driver://user:pass@localhost/dbname
```
mkangia, line 63: is it something that would need to be updated for production environments?

hq_superset/tests/config_for_tests.py
```diff
@@ -16,17 +16,21 @@
...
+    '0fXurIGyQM4HQYoe7feuwV8c1Kz_88BdmCNutLKiO38=',  # Don't reuse this!
+]
 
 # Any other additional roles to be assigned to the user on top of the base role
 # Note: by design we cannot use AUTH_USER_REGISTRATION_ROLE to
 # specify more than one role
 AUTH_USER_ADDITIONAL_ROLES = ["sql_lab"]
 
-HQ_DATA_DB = "postgresql://commcarehq:commcarehq@localhost:5432/test_superset_hq"
+HQ_DATABASE_URI = "postgresql://commcarehq:commcarehq@localhost:5432/test_superset_hq"
```
mkangia, line 33: nit: May be we need to update README to add note to add this DB to run tests.

hq_superset/utils.py
```diff
@@ -189,3 +226,53 @@ def array_is_falsy(array_values):
...
+    """
+    cast_functions = {
+        # 'BIGINT': int,
+        # 'TEXT': str,
+        'TIMESTAMP': partial(js_to_py_datetime, preserve_tz=False),
+        # TODO: What else?
+    }
+
+    column_types = {c.name: str(c.type) for c in table.columns}
+    for row in data:
```
mkangia, line 269: nit: probably its just me, though when I see `row`, I assume its a list. But here, this is a dictionary. That got me.

mkangia (commented):
Hey @kaapstorm 

Thanks for the work on this. This is a tricky one.
Nice refactor and clean up to make the repo better in general.

It was a big one so a lot of comments, mostly minor.

Just one critical query about why we chose to use OAuth instead of token based authentication to get updates from CommCareHQ.

hq_superset/services.py
kaapstorm: ba6286d

hq_superset/services.py
kaapstorm: 9f4b053

hq_superset/models.py
```diff
@@ -32,3 +45,71 @@ def update_dataset(self):
...
+        db.session.commit()
+
+    @classmethod
+    def get_by_domain(cls, domain):
+        return db.session.query(HQClient).filter_by(domain=domain).first()
+
+    @classmethod
+    def create_domain_client(cls, domain: str):
+        alphabet = string.ascii_letters + string.digits
+        client_secret = ''.join(secrets.choice(alphabet) for i in range(64))
```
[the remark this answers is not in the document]
kaapstorm: d30cc79

hq_superset/api.py
kaapstorm: This endpoint will be hit for every change for a doc_id, where a doc_id originally refers to a case or a form. Sometimes one doc_id can result in multiple rows, e.g. a data source definition that pulls out a question inside a repeat group.

hq_superset/models.py
kaapstorm: When I rebased (before reading this comment) I renamed this class "OAuth2Client", in line with its mixin. I did the same with "Token".

"HQOAuthClient" may have been a better name, considering it's got a "domain" field, but I feel like "OAuth2Client" is close enough, and better than "CommCareHQOAuth2Client".

hq_superset/utils.py
kaapstorm: That's what the type hint on line 256 is for.

While "row" could be a list, or a tuple, or a dictionary, "data" could be absolutely anything! Any variable contains data, doesn't it? And "table" is also ambiguous -- SQLAlchemy documentation uses the variable name "table" to refer to two or three different types.

So type hints, _in this specific case_, are helpful. The alternative would be to rename "data" to something more accurate, but less friendly, like "list_of_column_value_pairs", but then you wouldn't get any validation from your IDE.

Fun exercise: Open this in your IDE. Ctrl-click on the "columns" property of "table". Magic, right? Now delete the type hint, and try again.

kaapstorm: (You kinda handed me my favorite drum, and I just went ahead and banged it. :joy: :speak_no_evil:  Sorry about that.)

hq_superset/tests/config_for_tests.py
kaapstorm: You don't need to add it to run tests. `get_hq_database()` creates it automatically if it is missing.

hq_superset/migrations/alembic.ini
kaapstorm: Production environments use `superset db upgrade` to run migrations, and that should use [env.py to set it automatically](https://github.com/dimagi/commcare-analytics/blob/875ff55342997eb94e7dfc76fb3a4ebcaa862129/hq_superset/migrations/env.py#L17) for that environment.

hq_superset/migrations/README
kaapstorm: It was generated by Alembic. We don't really need it. 0974df5

README.md
kaapstorm: This [gets done automatically](https://github.com/dimagi/commcare-analytics/blob/d30cc792ecd236157356504bda5270fd9430a461/hq_superset/utils.py#L47) now by `get_hq_database()`.

hq_superset/api.py
kaapstorm: f0276dd

hq_superset/models.py
```diff
@@ -0,0 +1,34 @@
...
+        table = sqla_table.get_sqla_table_object()
+
+        with (
+            database.get_sqla_engine_with_context() as engine,
+            engine.connect() as connection,
+            connection.begin()  # Commit on leaving context
+        ):
+            delete_stmt = table.delete().where(table.c.doc_id == self.doc_id)
+            connection.execute(delete_stmt)
+            if self.data:
```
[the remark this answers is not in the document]
kaapstorm: We are addressing deletions. When all the rows for a form or a case are deleted, then `data` will be an empty list.

hq_superset/models.py
```diff
@@ -0,0 +1,34 @@
...
+    data_source_id: str
+    doc_id: str
+    data: list[dict[str, Any]]
+
+    def update_dataset(self):
+        database = get_hq_database()
+        try:
+            sqla_table = next((
+                table for table in database.tables
+                if table.table_name == self.data_source_id
```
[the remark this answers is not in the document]
kaapstorm: Yes. In fact, the table belongs to a schema which includes the domain name. That happens [when the table is created here](https://github.com/dimagi/commcare-analytics/blob/ba6286d124b483fe03eb527e0971cfa59e8502b0/hq_superset/services.py#L90). So the full name of the table for data source "abc123" in domain "foo" will be "hqdomain_foo.abc123".

hq_superset/hq_domain.py
```diff
@@ -26,6 +26,7 @@ def after_request_hook(response):
     "AuthDBView.login",
     "SelectDomainView.list",
     "SelectDomainView.select",
+    "DataSetChangeAPI.post_dataset_change",
```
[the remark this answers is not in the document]
kaapstorm: aedcb42

kaapstorm:
> one critical query about why we chose to use OAuth instead of token based authentication

OAuth 2.0 is token-based authentication. ([More info here](https://www.oauth.com/oauth2-servers/access-tokens/).) CommCare HQ already supports OAuth 2.0 token-based authentication for repeaters.

Do you have a different kind of token-based authentication in mind? What standard are you thinking of?

hq_superset/models.py
Charl1996: > I feel like "OAuth2Client" is close enough

Perfect!

README.md
mkangia: Oh, isn't this part about changing the auth_type to enable login via admin which is still applicable?

mkangia:
> Do you have a different kind of token-based authentication in mind? What standard are you thinking of?

Resolved offline.
OAuth is the better choice keeping security in mind and the library is already available via superset/flask so no new dependency was added.

README.md
kaapstorm: Good point! We don't need to log in as admin to create the database any more, but it is useful to know how to log in as admin if there is any other reason to do so. I'll add that back.

hq_superset/utils.py
mkangia: I kinda new what I was getting back when i asked that question, I knew you would find your way to typehints.

README.md
kaapstorm: af4878f

README.md
```diff
@@ -102,8 +102,19 @@ server. (Yes, it's "OAUTHLIB" this time, not "AUTHLIB" as before.)
...
 
+### Logging in as a local admin user
+
+There might be situations where you need to log into Superset as a local
+admin user, for example, to add a database connection. To enable local
+user authentication, in `superset_config.py`, set
+`AUTH_TYPE = AUTH_DB`.
+
+To return to allowing CommCare HQ users to log in, set it back to
+`AUTH_TYPE = AUTH_OAUTH`.
```
mkangia, line 114: nit: this can't be done in a production environment right? This is only for local setup?

hq_superset/api.py
```diff
@@ -12,6 +12,27 @@
 )
 
 from .models import DataSetChange
+from .oauth2_server import authorization, require_oauth
+
+
+class OAuth(BaseApi):
```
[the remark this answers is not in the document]
mkangia: this was taken offline.

hq_superset/models.py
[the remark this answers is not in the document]
mkangia: thanks, can we document that somewhere in this function?

mkangia:
New commits look good. Will wait for responses on other comments from the last review.

README.md
kaapstorm: :thinking: Doing this in a production environment will prevent CommCare HQ users from logging in. But for self-hosted environments it could be useful for maintenance or setting up additional databases, etc.

Do you think I should add more context? Or maybe give advice on when to use CommCare Ansible, and the rare cases when it might be appropriate to edit the config directly?

mkangia: >  Doing this in a production environment will prevent CommCare HQ users from logging in

I believe that is true hence this should not be used for production environments.

I was very confused when I saw this for the first time. For me, I have always thought of this as something to be used only locally.
Though now when you mention maintenance, I see there is no way to access UI for an admin without this change, so it is the only way out.

> the rare cases when it might be appropriate to edit the config directly

Right, we should just caution that doing this change disables usual new user login (I assume the ones who are logged in will continue to work), so, on a production environment this should be only done when system isn't in use.

P.S: why do we need this change though? I comment about that would be super helpful in the code

hq_superset/models.py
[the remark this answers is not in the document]
kaapstorm: 2c0c766

README.md
kaapstorm: ceb8f80

mkangia:
New commits look good :+1:

mkangia:
Hey @kaapstorm 

I see some open comments from my earlier review. Waiting on those to do a full pass again. I hope these will be quick since they are mostly minor.

Charl1996 (approved):
Did a brief review of the newer commits and all seems fine. I'm happy with this PR.

hq_superset/services.py
```diff
@@ -143,6 +152,48 @@ def to_sql(df, replace=False):
...
 
 
+def subscribe_to_hq_datasource(domain, datasource_id):
+    client = _get_or_create_oauth2client(domain)
+    hq_request = HQRequest(url=datasource_subscribe(domain, datasource_id))
+    webhook_url = current_app.url_for(
+        'DataSetChangeAPI.post_dataset_change',
+        _external=True,
+    )
+    token_url = current_app.url_for('OAuth.issue_access_token', _external=True)
```
Charl1996, line 162: (related to [this](https://github.com/dimagi/commcare-hq/pull/34179#issuecomment-1985185551) comment)

[This](https://stackoverflow.com/questions/66663507/flask-url-for-http-instead-of-https) is an interesting post. It seems like one might need to explicitly specify the `https` scheme should be used.

hq_superset/oauth2_server.py
```diff
@@ -39,6 +40,10 @@ def save_token(token, request):
 
 
 def config_oauth2(app):
+    authlib_logger = logging.getLogger('authlib')
+    authlib_logger.addHandler(logging.StreamHandler(sys.stdout))
+    authlib_logger.setLevel(logging.DEBUG)
```
mkangia, line 45: :spiral_notepad:

kaapstorm:
49d575aa47411ba211e1dda3ec260600c6dd64c4 fixed the problem with syncing from CommCare Staging. I don't like it, but the [right way to do it](https://stackoverflow.com/a/45333882) didn't work.

"If at first (or second, or third, or fourth) you don't succeed, use a bigger hammer."

hq_superset/services.py
kaapstorm: 9042fb6e37d1e5a2f9d9a19e264f62c615ce09fd

README.md
kaapstorm: I think we can definitely pull CommCare Analytics into the current discussion around UCR feature flags.

kaapstorm:
I think that's everything. If there is anything I've missed, please link to it.

hq_superset/__init__.py
```diff
@@ -22,6 +22,13 @@ def flask_app_mutator(app):
     app.strict_slashes = False
     override_jinja2_template_loader(app)
 
+    # A proxy (maybe) is changing the URL scheme from "https" to "http"
+    # on commcare-analytics-staging.dimagi.com, which breaks the OAuth
+    # 2.0 secure transport check despite transport being over HTTPS. I
+    # hate to do this, but werkzeug.contrib.fixers.ProxyFix didn't fix
+    # it. So I've run out of better options. (Norman 2024-03-13)
+    os.environ['AUTHLIB_INSECURE_TRANSPORT'] = '1'
```
mkangia, line 30: Hey @kaapstorm 

Good find on the proxy redirection.

I can make peace with this change to unblock QA/UAT.

It's a positive that this is on analytics server that we have control/access to and not superset.

I was trying to find where in nginx conf, this is getting set and compare that with what we have for HQ though HQ nginx template has a lot happening.

Couple of things I noticed that could be faulty
1. we have two "server" blocks with the same server name in nginx conf. I am not sure if that is expected/correct
2. I don't know what the following block does
```
if (!-f $request_filename) {
      proxy_pass http://superset;
      break;
    }
```

@sravfeyn assuming you have more insight when this was setup.
Can you see something faulty with the current nginx conf or point out what could redirecting urls to http?
I don't know what this block does

kaapstorm: I took a quick look at the nginx config, and realised it will take me a bit longer to change and test it. I'll merge this PR, and I'll create a follow up ticket to see how we can configure nginx so that we can use werkzeug's "ProxyFix" class (or not have to use it at all) instead of using this environment variable.

mkangia: Okay @kaapstorm, thanks for taking a look.

I am fine with this merge till this stays on staging and is not released to production environment until the secure fix is done.

PR #42, opened 2024-02-28
Prefactor for background sync

kaapstorm:
This PR separates out the prefactor commits for PR #41 

These commits do not make any significant change to functionality. They just clean up code a little.

Easiest to review by commit. :blowfish: :tropical_fish: :fish: :shark: :dolphin:

### DIFF
--- a/hq_superset/utils.py
+++ b/hq_superset/utils.py
@@ -16,6 +16,10 @@ HQ_DB_CONNECTION_NAME = "HQ Data"
 ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES = 5_000_000  # ~5MB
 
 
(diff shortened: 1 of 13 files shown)


hq_superset/utils.py
```diff
@@ -16,6 +16,10 @@
 ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES = 5_000_000  # ~5MB
 
 
+class CCHQApiException(Exception):
```
Charl1996: +1 for this change, although what do you think about adding this to an `exceptions.py` file? I don't have strong feelings either way, just wanting to gauge your thoughts...

Charl1996 (approved):
Nice! Thanks for this!

PR #45, opened 2024-05-29
Delete downloaded files in case of exception as well

mkangia:
https://dimagi.atlassian.net/browse/SC-3646

We were retaining files in case the task fails.

Also reached out to @sravfeyn to confirm this change does not cause unexpected issues [here](https://github.com/dimagi/commcare-analytics/pull/28/files#r1618328156)

The other [place](https://github.com/dimagi/commcare-analytics/blob/bd46e6e7d590879c7934224d2d729227c5fe0b04/hq_superset/services.py#L155) mentioned in the [ticket](https://dimagi.atlassian.net/browse/SC-3646?focusedCommentId=328631) should be handled by this change itself because the exception raised is caught in `refresh_hq_datasource_task` after that.

### DIFF
--- a/hq_superset/tasks.py
+++ b/hq_superset/tasks.py
@@ -12,4 +12,6 @@ def refresh_hq_datasource_task(domain, datasource_id, display_name, export_path,
     except Exception:
         AsyncImportHelper(domain, datasource_id).mark_as_complete()
         raise
-    os.remove(export_path)
+    finally:
+        if os.path.exists(export_path):
+            os.remove(export_path)
--- a/hq_superset/tests/test_views.py
+++ b/hq_superset/tests/test_views.py
@@ -231,6 +231,26 @@ class TestViews(HQDBTestCase):
                 'ds1'
             )
 
+    def test_trigger_datasource_refresh_with_errors(self, *args):
+        from hq_superset.views import (
+            ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES,
+            trigger_datasource_refresh,
+        )
+
+        file_path = '/file_path/towards/dimagi'
+        file_size = ASYNC_DATASOURCE_IMPORT_LIMIT_IN_BYTES - 1
+        with (
+            patch("hq_superset.views.download_and_subscribe_to_datasource", return_value=(file_path, file_size)),
+            patch("hq_superset.views.get_datasource_defn", return_value=TEST_DATASOURCE),
+            patch("hq_superset.views.refresh_hq_datasource", side_effect=Exception('mocked error')),
+            patch('hq_superset.views.os.remove') as os_remove_mock
+        ):
+            response = trigger_datasource_refresh('test1', 1, 'ds_name')
+
+            self.assertEqual(response.status_code, 302)
+            self.assertEqual(response.location, "/tablemodelview/list/")
+            os_remove_mock.assert_called_once_with(file_path)
+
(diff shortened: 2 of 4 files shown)


hq_superset/tasks.py
```diff
@@ -11,5 +11,12 @@ def refresh_hq_datasource_task(domain, datasource_id, display_name, export_path,
         refresh_hq_datasource(domain, datasource_id, display_name, export_path, datasource_defn, user_id)
     except Exception:
         AsyncImportHelper(domain, datasource_id).mark_as_complete()
+        _remove_export_file(export_path)
         raise
-    os.remove(export_path)
+    else:
```
ajeety4, line 16: nit: should we consider using `finally` instead of `else` ?

ajeety4 (approved):
:+1: 
Looks good. Just left a non-blocking comment.

hq_superset/tasks.py
mkangia: I was not sure finally would work since we are raising exception within except block again. let me check

mkangia: That did work, let me update.

mkangia: https://github.com/dimagi/commcare-analytics/pull/45/commits/a5afef5c56e26d5491324e0811943941c290a0f5

mkangia: took the opportunity to simplify and merge those commits. Overall change remains the same.

ajeety4:
Hey @mkangia 
Just realised , there is one more place that the file may remain if exceptions occurs for data source refresh in sync mode [here](https://github.com/dimagi/commcare-analytics/blob/a5afef5c56e26d5491324e0811943941c290a0f5/hq_superset/views.py#L130)

mkangia:
Ajeet, I rebased the feedback fix since this was a very small PR. Requesting for a quick pass on review again. The overall changes remain the same as you reviewed.

mkangia:
oh yes, good find. Addressed that as well. Trying to add a test for it as well.

hq_superset/tasks.py
```diff
@@ -12,4 +12,6 @@ def refresh_hq_datasource_task(domain, datasource_id, display_name, export_path,
     except Exception:
         AsyncImportHelper(domain, datasource_id).mark_as_complete()
         raise
-    os.remove(export_path)
+    finally:
```
sravfeyn, line 15: 👍

hq_superset/views.py
```diff
@@ -127,10 +127,18 @@ def trigger_datasource_refresh(domain, datasource_id, display_name):
...
+                domain, datasource_id, display_name, path, datasource_defn, None
+            )
+        except Exception:
+            flash(
+                "The datasource refresh failed. "
+                "Please try again or report if issue persists.",
+                "danger",
+            )
+        finally:
+            os.remove(path)
```
sravfeyn, line 141: 👍

sravfeyn (approved):
LGTM

ajeety4 (approved):

mkangia:
fyi @ajeety4 
test added for error handling.

hq_superset/tests/test_views.py
```diff
@@ -231,6 +231,26 @@ def test_datasource_upload(self, *args):
                 'ds1'
             )
 
+    def test_trigger_datasource_refresh_with_errors(self, *args):
```
ajeety4, line 234: :+1:

ajeety4 (approved):

PR #49, opened 2024-06-27
Return 404 on missing table

Charl1996:
This PR simply returns a 404 status code with a message saying the data source is missing when the data source is removed on CCA but the repeater still sends updates to CCA. 

There's a spec open for review for properly handling removed data sources on CCA and its corresponding repeater on HQ.

### DIFF
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -10,6 +10,7 @@ from superset.views.base import (
     json_error_response,
     json_success,
 )
+from .exceptions import TableMissing
 
 from .models import DataSetChange
 from .oauth2_server import authorization, require_oauth
@@ -66,4 +67,9 @@ class DataSetChangeAPI(BaseApi):
             return json_error_response(
                 'Invalid JSON syntax',
                 status=HTTPStatus.BAD_REQUEST.value,
+            )
+        except TableMissing:
+            return json_error_response(
+                'Data source not found',
+                status=HTTPStatus.HTTP_404_NOT_FOUND.value,
             )


hq_superset/api.py
```diff
@@ -67,3 +68,8 @@ def post_dataset_change(self) -> FlaskResponse:
                 'Invalid JSON syntax',
                 status=HTTPStatus.BAD_REQUEST.value,
             )
+        except TableMissing:
+            return json_error_response(
+                'Data source not found',
+                status=HTTPStatus.HTTP_404_NOT_FOUND.value,
```
mkangia, line 74: will this cause the repeater to constantly keep trying? in that case this won't be better than what this currently is.

Charl1996: It won't stop the repeater from retrying, but I thought this change at least gives more info w.r.t why the request is failing. Ideally this won't happen after we properly handle data source table removals / repeater removals, but having this feels a little better than is.

That said, I'm willing to let this go, but let's hear what @kaapstorm says.

ajeety4: On second thought, I think this is still good to have irrespective of how it is being used in external service as a good REST practice.

ajeety4 (approved):

hq_superset/api.py
kaapstorm: Yeah, I second @ajeety4 's second thought. If a sender (even an API user) expects a thing to be there, and it isn't there, a 404 response sounds right to me.

What's more, the alternative is a 500, which suggests that the problem lies with the server, not the sender, and that sounds wrong. The sender should be told that they are trying to update a DataSet that isn't there.

kaapstorm (approved):

hq_superset/api.py
mkangia: Okay :+1:

PR #54, opened 2024-08-21
Make data source update async

mkangia:
In order to make the update dataset request as fast as possible so it can keep up with the load of requests received from HQ, move the actual update to a background task via celery

### DIFF
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -60,16 +60,19 @@ class DataSetChangeAPI(BaseApi):
 
         try:
             request_json = json.loads(request.get_data(as_text=True))
-            change = DataSetChange(**request_json)
-            change.update_dataset()
-            return json_success('Dataset updated')
         except json.JSONDecodeError:
             return json_error_response(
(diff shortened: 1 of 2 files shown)


hq_superset/api.py
```diff
@@ -60,16 +60,20 @@ def post_dataset_change(self) -> FlaskResponse:
...
+        except:
             return json_error_response(
-                'Data source not found',
-                status=HTTPStatus.HTTP_404_NOT_FOUND.value,
+                'Could not parse change request',
+                status=HTTPStatus.BAD_REQUEST.value,
             )
+
+        process_dataset_change.delay(request_json)
+        return json_success('Dataset change accepted')
```
kaapstorm, line 79: There is a status code for this :)

```suggestion
        return json_success(
            'Dataset change accepted'
            status=HTTPStatus.ACCEPTED.value,
        )
```

kaapstorm (approved):
This is excellent. My suggestion is not blocking.

kaapstorm (approved):

PR #56, opened 2024-08-22
Skip revoking tokens

mkangia:
Each time a CommCareHQ repeater sends a request to CCA, it requests a new token.
On CCA, when we were generating a new token, we would revoke all old ones for that client.

In case of concurrent requests by repeaters one new token revokes the other one yet to be used, which lead to authorization errors.
So,
we are skipping revoking tokens when generating new tokens.
The expiry time was removed as a part of this investigation and is now added back but with a much shorter time of 10 seconds.

As a follow up, 
we will clean up expired tokens via a celery task and/or not need to get a new token for each repeat record.

### DIFF
--- a/hq_superset/models.py
+++ b/hq_superset/models.py
@@ -1,4 +1,3 @@
-import time
 from dataclasses import dataclass
 from typing import Any
 
@@ -7,7 +6,6 @@ from authlib.integrations.sqla_oauth2 import (
     OAuth2TokenMixin,
 )
 from cryptography.fernet import MultiFernet
-from sqlalchemy import update
 from superset import db
 
 from .const import OAUTH2_DATABASE_NAME
@@ -77,17 +75,6 @@ class OAuth2Client(db.Model, OAuth2ClientMixin):
 
     def check_client_secret(self, plaintext):
         return self.get_client_secret() == plaintext
-
-    def revoke_tokens(self):
-        revoked_at = int(time.time())
-        stmt = (
-            update(OAuth2Token)
-            .where(OAuth2Token.client_id == self.client_id)
-            .where(OAuth2Token.access_token_revoked_at == 0)
-            .values(access_token_revoked_at=revoked_at)
-        )
-        db.session.execute(stmt)
-        db.session.commit()
 
 
 class OAuth2Token(db.Model, OAuth2TokenMixin):
--- a/hq_superset/oauth2_server.py
+++ b/hq_superset/oauth2_server.py
@@ -41,7 +41,7 @@ require_oauth = ResourceProtector()
 def config_oauth2(app):
     authlib_logger = logging.getLogger('authlib')
     authlib_logger.addHandler(logging.StreamHandler(sys.stdout))
-    authlib_logger.setLevel(logging.DEBUG)
+    authlib_logger.setLevel(logging.INFO)
 
     authorization.init_app(app)
     authorization.register_grant(grants.ClientCredentialsGrant)
--- a/hq_superset/oauth2_server.py
+++ b/hq_superset/oauth2_server.py
@@ -19,12 +19,13 @@ from .models import OAuth2Client, OAuth2Token, db
 def save_token(token: dict, request: FlaskOAuth2Request) -> None:
     client = request.client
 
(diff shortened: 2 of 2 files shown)


mkangia:
Hey @kaapstorm 

Requesting for a review on this.

I would like to deploy this on staging only for now and see how it goes.

kaapstorm (approved):
Approving for the sake of experimenting. But I worry that this is answering the wrong question.

If we don't revoke expired tokens, then tokens effectively never expire because they would be valid forever.

So (for now) why don't we rather just issue tokens that expire in, say, 12 months? Or a random period between 6 and 12 months? That would have the same outcome -- (after the initial refresh) they don't get revoked.

What Google does is set a short expiry for the first token, and set a long (like 6 months) expiry for refresh tokens. Maybe (at some point in the future, not now) we could do something like that, where the first token is only used for the initial import, and the refresh token is used for syncing? I'm not sure. But I think if we set expiry to 12 months it gives us some time to think about it.

mkangia:
Noted @kaapstorm 

I hear you. I will definitely add some value for `expires_in` before releasing this to production.
Currently not expiring at all from CCA end to keep it straightforward.

As of now, will just deploy to staging to test. And then will also test with expiry.

mkangia:
Hey @kaapstorm 

I changed my approach. Since each request from repeater, takes a new token, the expiry time is redundant so I set it to the smallest possible value which should not cause any trouble.

Description is updated as well.

I tested this on staging and everything went well.

This should be good to deploy to production now.
The follow up can be taken at its own time and is not a blocker for this.

kaapstorm:
> Since each request from repeater, takes a new token

Really?! HQ is supposed to save the last token in `ConnectionSettings.last_token_aes`. Maybe it's not saving it. Or maybe it's not using the token it saved. :thinking:  :unamused: 

What happens if we set `expires_in=365 * 24 * 60 * 60` ?

kaapstorm:
I wonder if this is related: https://dimagi.sentry.io/issues/5746233109/?project=136860

kaapstorm (changes requested):
On further consideration, I don't think this is a good idea. I think we should rather fix the way that tokens are managed in CommCare than break the way that tokens are managed in CCA, even temporarily.

If fixing the way that tokens are managed in CommCare turns out to be difficult, we can revisit this approach.

mkangia:
> Really?! HQ is supposed to save the last token in ConnectionSettings.last_token_aes. Maybe it's not saving it. Or maybe it's not using the token it saved

It seems to be and I believe that has been the core of misunderstanding here.  Do you know where I can look to confirm?


> What happens if we set expires_in=365 * 24 * 60 * 60 ?

As for how things are now, the token won't ever be used again and will stay active for one year.
Even when, expires_in was 0, new tokens were being requested.

mkangia:
Hey @kaapstorm 

> than break the way that tokens are managed in CCA

Just to be clear about what changes here.

As of now any old tokens for the client stay in DB and are marked as revoked.

After this change, the tokens wont be marked as revoked and will expire after 10 seconds.

I don't think that change is making it worse. Let me know if you think otherwise.

Additionally, to support the change here,
And irrespective of the motivation here, it does seem its better to expire tokens rather than manually revoking them unless its a requirement of OAuth that at once, only one token should be valid for a client.

kaapstorm (approved):

mkangia:
Releasing this on staging first

mkangia:
Tested on staging, things working fine
1. New login
2. Refresh
3. Repeaters from HQ

Deployed to production as well.

mkangia:
Hey @kaapstorm 

> I'm not sure what you mean by "manually" here

I was referring to this task that I removed in this [commit](https://github.com/dimagi/commcare-analytics/pull/56/commits/4c73ffd45200a43326a72795e15e7530eb349196#diff-2f0793ff3f61bb533690fde0611fc67279c392df6f0ca70a5a7ba6b8b2c4769dL21)

> Based on that definition, it sounds like CCA doesn't have to revoke an old token when a new token is requested. (In other words, we're good.)

Ah! Very nice. That gives me more confidence in our approach here.

PR #58, opened 2024-08-28
Assign HQ-defined roles to CCA user 

Charl1996:
Putting this out here so folks can start taking a look. I still need to clean up the commits a bit.

I'm also now looking into fetching the CCA roles from the oauth scope instead of an API endpoint.


Charl1996:
Now onto cleaning up commit history...

Charl1996:
Closing because I'm creating a new PR with a slightly better commit history

PR #59, opened 2024-08-29
Assign HQ-defined roles to CCA user

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/SC-3473)

This PR expands on the DomainSyncUtil class's `sync_domain_role` method to now consider the roles for a particular user as defined on CommCare HQ. [This PR](https://github.com/dimagi/commcare-hq/pull/34987) is for returning the CCA roles for a particular user.

Note: This work will go [through QA](https://dimagi.atlassian.net/browse/QA-7055) and is already on staging.

### Approach
The approach is slightly different than the [spec](https://docs.google.com/document/d/1WTg8T37plRdjOK7wXl3-yip5SF_qG4LiD4hzMj57XOI/edit#heading=h.p76anzyemt26) suggests in the following ways:
1. We now create only 1 role for a user _per domain_ which has the correct permissions defined as per CommCare HQ. The role's name is composed [here](https://github.com/dimagi/commcare-analytics/blob/816f1f1735439eaf797021e25e6c58efcbdec726/hq_superset/utils.py#L289-L291) and is identifiable by domain and user.
2. When the user's permissions change on HQ and the user logs in again, the permissions for the role is overwritten with the new permissions.

By having only 1 role per domain user and setting the permissions for that role we won't end up having redundant roles that we'll need to clean up.

#### Side note
Since we're not assigning the `Gamma` by default to all users anymore (it's a configurable role on HQ now), it was found that the CCA user cannot log in to Superset since the privilege of having a profile on Superset is a permission the user needs.    If the user does not have this permission, the user's browser will redirect endlessly resulting in a browser error.

To account for this a new standard role, namely `hq_user`, is assigned to all CCA users, regardless of their permissions on CommCare HQ. This role has the necessary ("baseline") permissions needed for any users to be able have a profile on Superset/ be able to log in to Superset.

This means that users will need at least 3 distinct roles if they want to be able to access a domain's data on Superset:
1. HQ user role (new): The role that grants the user permission to be able to have a profile on Superset.
2. Domain schema role (existed before this PR): This role grants (and restricts) the user's access to a particular domain.
3. Domain user role (new): This role contains the necessary configurable permissions as defined on CommCare HQ and governs how the user can behave on the particular domain.

In addition to the above roles the user is able to have more roles if so defined on CommCare HQ. The roles available on HQ is currently `Gamma`, `dataset_editor` and `sql_lab`.


mkangia:
Hey @Charl1996 

Can you link the related HQ PR in the description please?
https://github.com/dimagi/commcare-hq/pull/34987

Charl1996:
> Hey @Charl1996
> 
> Can you link the related HQ PR in the description please? [dimagi/commcare-hq#34987](https://github.com/dimagi/commcare-hq/pull/34987)

Done

superset_config.example.py
```diff
@@ -70,11 +70,6 @@
 # The default user self registration role
 AUTH_USER_REGISTRATION_ROLE = "Gamma"
 
-# Any other additional roles to be assigned to the user on top of the base role
-# Note: by design we cannot use AUTH_USER_REGISTRATION_ROLE to
-# specify more than one role
-AUTH_USER_ADDITIONAL_ROLES = ["sql_lab"]
```
mkangia, line 76: Hey @Charl1996 

How do we roll this change out? Is there anything specifically needed for existing users? or their permission would reset correctly when they login.
So, will this role get removed from their account automatically now, if they had it earlier?

How does platform role work along with domain role? If someone gets the role sql_lab, it would be applicable only for the project they are currently on? and when they switch project, their platform roles are fetched and updated?

mkangia (commented):
Beautiful PR @Charl1996 

I enjoyed reviewing this. :100: 

Are we taking this through QA?

Charl1996:
> Are we taking this through QA?

Yes. I updated the description to make it more visible.

hq_superset/oauth.py
```diff
@@ -56,6 +56,15 @@ def set_oauth_session(self, provider, oauth_response):
         # }
         session[SESSION_OAUTH_RESPONSE_KEY] = oauth_response
 
+    def set_role_permissions(self, role, permissions):
```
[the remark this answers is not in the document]
Charl1996: I was also thinking it feels a little out of place, but then all the permissions-related functionality sits in the class this class inherits from, and since this is a permissions-related method I thought it's probably best putting it here. Any other suggestions are welcome.

hq_superset/const.py
```diff
@@ -2,3 +2,67 @@
...
+CAN_DELETE_PERMISSIONS = "can_delete"
+
+WRITE_PERMISSIONS = [
+    CAN_WRITE_PERMISSION,
+    CAN_EDIT_PERMISSION,
+    CAN_ADD_PERMISSION,
+    CAN_DELETE_PERMISSIONS,
+]
+
+READ_ONLY_MENU_PERMISSIONS = {
```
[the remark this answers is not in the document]
Charl1996: Correct. I played around a lot with how to best separate the read and write whilst keeping duplication to a minimum, but this current way of doing it feels like the most granular and configurable approach. You can set specific permissions for specific view/menu's

superset_config.example.py
Charl1996: > or their permission would reset correctly when they login.

Yes, when they log in their permissions will reset correctly.

> So, will this role get removed from their account automatically now, if they had it earlier?

Correct. My thoughts are that if a user needs access they need to now have it explicitly granted on HQ (even if it means they might not have that permissions until it's explicitly given to them on HQ if they had it before), but maybe I can check with AEs as well.

> How does platform role work along with domain role? If someone gets the role sql_lab, it would be applicable only for the project they are currently on? 

Ah! Brilliant question! See response below.

> and when they switch project, their platform roles are fetched and updated?

More accurately, fetched and _overwritten_.

With the way it works in this PR, when a user logs in their user on CCA will be assigned roles/permissions based on only the domain's configuration they logged in to, so if their user on that domain on HQ does not have the Gamma role for instance, the user on CCA will not be granted that role when logging in. This is true even though the user might have Gamma on another domain; when they log into that other domain they will be given the Gamma role. So even though the platform role is technically platform wide, the user will only ever be able to have that role on the domain which gives it to the user. This is true because every time a user logs in we overwrite the user's roles completely with the appropriate roles for the domain the user logs into.

mkangia: > Correct. My thoughts are that if a user needs access they need to now have it explicitly granted on HQ (even if it means they might not have that permissions until it's explicitly given to them on HQ if they had it before), but maybe I can check with AEs as well.

Right. It seems if we roll this out, people would lose access, at least temporarily, till they are assigned the permission on HQ.
The good thing is that the permissions setup is already available on HQ. So we can ask projects to set them even before we release this.

Additionally, we should be explicit that the permissions are updated only at login and when switching domains so they need to re-login to see the updated permissions from HQ.

Charl1996: > permissions are updated only at login

This is not entirely true; the permissions are updated when selecting a domain, so even though the user is "logged in" to superset, when they change domains the permissions will update (sorry, maybe I wasn't too clear on this distinction in my previous comment)

mkangia: Right, that is what I meant by "only at login and when switching domains"?

may be a correction to that "they need to re-login or re-select their domain" (I don't know if they can do that, if they have only one domain, which would be for most users?)

Charl1996: Oh! Sorry, yeah, now I understand what you meant. You're right. Although there is a way when logged into the app to go to the "select domains" page without having to explicitly log out and re-auth.

Charl1996: I've checked with AEs and it's OK that users might loose temporary access if their HQ permissions isn't set up correctly; we just have to make sure to communicate this change when deploying the new changes.

mkangia: Cool.

mkangia (approved):
looking forward to QA on this.

hq_superset/hq_url.py
```diff
@@ -32,5 +32,5 @@ def datasource_unsubscribe(domain, datasource_id):
     )
 
 
-def user_domain_roles(username, domain):
```
Charl1996: HQ will now get the user from the request.

Charl1996:
QA passed on this.

mkangia (approved):

kaapstorm (approved):

kaapstorm (approved):

mkangia (approved):

Charl1996:
Please note that an [update](https://github.com/dimagi/commcare-analytics/pull/59/commits/cb57fa67dbc46e9b0bb9b8f9aeb52a3056914bc3) has been made which basically does the following:
1. If a HQ user has "edit" permissions set for CCA, assign the superset `Gamma` role
2. If a HQ user does not have "view" or "edit" permissions set for CCA, the user cannot access CCA.

hq_superset/tests/test_utils.py
```diff
@@ -24,3 +25,90 @@ def test_doctests():
...
+        self._ensure_platform_roles_exist(security_manager)
+
+        get_domain_access_mock.return_value = self._to_permissions_response(
+            can_write=True,
+            can_read=True,
+            roles=[],
+        )
+        additional_roles = DomainSyncUtil(security_manager)._get_additional_user_roles("test-domain")
+        assert len(additional_roles) == 1
+        assert additional_roles[0].name == "Gamma"
```
mkangia, line 45: should not the user always have the custom domain role?

hq_superset/views.py
```diff
@@ -242,5 +246,11 @@ def select(self, hq_domain):
             )
             return redirect(url_for('SelectDomainView.list', next=request.url))
         response.set_cookie('hq_domain', hq_domain)
-        DomainSyncUtil(superset.appbuilder.sm).sync_domain_role(hq_domain)
+        if not DomainSyncUtil(superset.appbuilder.sm).sync_domain_role(hq_domain):
+            flash(
+                "You don't have the necessary HQ permissions to access this domain.",
```
mkangia, line 251: nit: might be useful to add the name of the domain to the message so the user knows which domain they were trying to access?

Charl1996: Good idea!

hq_superset/tests/test_utils.py
Charl1996: Not anymore, since we assign the Gamma role to "edit" users and will only create and assign a "read only" role to "read" users. 

I think I'm going to update the custom user domain role name such that it makes clear that it's a read-only role. So the name becomes `<domain>_user_<user_id>_read_only`

hq_superset/views.py
Charl1996: https://github.com/dimagi/commcare-analytics/pull/59/commits/f5384d1e50aa6c3b11a8b6913838e4168984c643

hq_superset/tests/test_utils.py
Charl1996: https://github.com/dimagi/commcare-analytics/pull/59/commits/4d52e10f48937c255080e71371ec499e52c26bf1

mkangia: Oh you are just testing for the additional roles, sorry.

I was thinking about the domain role that is added to the user so they have access to just their domain?
This one, https://github.com/dimagi/commcare-analytics/blob/d8feb1d42c01b8b00ea8cc11bf81ae706c309d01/hq_superset/utils.py#L123

mkangia: > I think I'm going to update the custom user domain role name such that it makes clear that it's a read-only role. So the name becomes <domain>_user_<user_id>_read_only

Is that in line with the spec? We wanted to keep the naming consistent with roles. 
Additionally, this adds the extra overhead of changing the name to be without "read_only" if the permissions are updated, which is best avoided.

hq_superset/utils.py
```diff
@@ -139,29 +188,109 @@ def _ensure_schema_created(domain):
...
+
+    def _get_additional_user_roles(self, domain):
+        domain_permissions, platform_roles_names = self._get_domain_access(domain)
+        if self._user_has_no_access(domain_permissions):
+            return []
+
+        additional_roles = []
+        if domain_permissions[CAN_WRITE_PERMISSION]:
+            platform_roles_names.append(
+                'gamma',
```
mkangia, line 211: should this be the one with "G"? The one on superset is "Gamma" or "gamma"?
Extracted it to a constant would be helpful to avoid confusion

hq_superset/utils.py
```diff
@@ -139,29 +188,109 @@ def _ensure_schema_created(domain):
...
+        for view_menu_name, permissions_names in menu_permissions.items():
+            menus_permissions.extend([
+                self.sm.add_permission_view_menu(permission_name, view_menu_name)
+                for permission_name in permissions_names
+            ])
+        return menus_permissions
+
+    @staticmethod
+    def _domain_user_role_name(domain, user):
+        return f"{domain}_user_{user.id}_read_only"
```
mkangia, line 293: Hey @Charl1996 

I wanted to check with you if its possible to add the HQ role UUID to the name?
While looking at the ticket to expire permissions after every hour, I realized we currently don't a way of linking HQ role to the CCA role, which could be useful.
If its not a big lift, would be good to do it now since we are releasing this now.

hq_superset/tests/test_utils.py
Charl1996: The domain role that gives a user access to only their own domain is still given as usual and have nothing to do with these changes.

> Is that in line with the spec? We wanted to keep the naming consistent with roles.

Maybe not, but I thought it's probably a good idea to be able to see what permissions the role gives, since some users will have the `user_:user_id:_read_only` role and others not (I'm thinking in the scenario where someone's looking at a user's roles and want to know what access they have). If we give it a more generic name one would need to go and view the permissions individually to see what the role means.

Let me know if I need to change it back or if this new way is OK.

> Additionally, this adds the extra overhead of changing the name to be without "read_only" if the permissions are updated, which is best avoided.

It's just a one-liner, but yeah, that would need to update if we ever need to give more than read-only access that we decide to put into this role (as opposed to creating a different role).

hq_superset/utils.py
Charl1996: > Extracted it to a constant would be helpful to avoid confusion

Good idea

hq_superset/utils.py
Charl1996: I don't think it would be that much of a big lift. I'll look into it.

hq_superset/tests/test_utils.py
mkangia: Hey @Charl1996 

> If we give it a more generic name one would need to go and view the permissions individually to see what the role means.

So, the roles like dataset_editor are different right? This one is just for read only and write access? So the role would either be read only or write?

When a user's permissions are upgraded from read only to write, do we change the name of the same role or a new one gets created & old one deleted?

> Let me know if I need to change it back or if this new way is OK.

I think its fine to keep it if you think its going to be useful on the UI. I am just unsure of the  benefit over the cost of the overhead it brings, though we can always revisit this later if needed.

hq_superset/utils.py
mkangia: Awesome, thank you!

hq_superset/tests/test_utils.py
mkangia: Actually,
Thinking more about this,
if the naming is updated to have `_read_only` in the role name, and if the users role is upgraded from view only to write, won't CCA end up having two roles when the syncing of role happens, since it will simply add a new role without the `_read_only` bit, leaving the old one around?

hq_superset/utils.py
Charl1996: @mkangia I had a look into this and hit a snag...the Admin role is a virtual role and is never persisted in the database, so if the user is admin we don't have a role ID that we could use.

Maybe we should rethink how we track the role linkage. I'm going to hold off on this for now.

mkangia: >  if the user is admin we don't have a role ID that we could use.

so the id could be "admin"?

>  I'm going to hold off on this for now.

We don't need to block on this. Fine to go ahead without the role id if needed.

mkangia:
All looks good here except this point that I believe will add redundant roles in CCA
https://github.com/dimagi/commcare-analytics/pull/59#discussion_r1822857924

hq_superset/tests/test_utils.py
Charl1996: @mkangia 

> When a user's permissions are upgraded from read only to write, do we change the name of the same role or a new one gets created & old one deleted?

We don't assign the role and assign the Gamma instead (leaving the old one lying around).

Now that you mention this, I'm beginning to think we should simplify this to follow the previous way where we simply assigned a user domain role which will contain all the permissions (read or write). The user will always have this role, but the permissions will only be different. 

In the case of an edit user we can simply assign all the gamma permissions to the role. This is as easy as
```
new_permissions = gamma_role.permissions
```
(why did I not think of this earlier?)

What do you think?

mkangia: I'd suggest to use the Gamma role itself.

> assign all the gamma permissions to the role

This seems like duplication of permissions.
Additionally, Superset/Flask architecture supports adding multiple roles, so seems good to leverage a role that already exists with necessary permissions instead of copying its permissions over.

Though, effectively both approaches have the same end result, so seems okay to do this if you feel that is better.

Its additionally about if there is any difference of effort in these two
1. assign gamma role permissions
2. remove gamma role from user if the user does not have edit access

Charl1996: I'd like to add that for an "edit" user we assign the gamma role which also gives all the view permissions, so we don't assign the domain user role (the role that specifies which view/edit permissions the user has on a domain), meaning if a user was a "read only" user and is upgraded to the "edit" user there's a redundant view-only role in the DB that we'd need to clean up.

Charl1996: See [this PR](https://github.com/dimagi/commcare-analytics/pull/71) for alternative approach.

mkangia: > if a user was a "read only" user and is upgraded to the "edit" user there's a redundant view-only role in the DB that we'd need to clean up

Oh yes, I missed that.

mkangia:
Hey @Charl1996 

Can you confirm what happens if HQ does not respond well or say responds with a 500 when fetching roles/permissions?
I had this happening locally and for some reason CCA continued to function which was unexpected.

mkangia:
Hey @Charl1996 
Due to recent api changes that broke the integration, I was getting 500 from HQ, locally, that is where I noticed the pattern.
I see how this is handled now. I need to add the same here, https://github.com/dimagi/commcare-analytics/pull/72#discussion_r1828027612

I think the error message is a bit misleading thought. May be we could update to tell the user that we could not sync their permission instead of saying that they don't have permission.
Thank you.

Charl1996:
If we merge [this PR](https://github.com/dimagi/commcare-hq/pull/35312) whilst keeping the resource on `v1` I'd need to just add the update here. I'll wait till it's merged before I do it.

hq_superset/utils.py
```diff
@@ -120,16 +129,57 @@ class DomainSyncUtil:
...
+        self.sm.get_session.add(current_user)
+        self.sm.get_session.commit()
+        return True
+
+    def _ensure_hq_user_role(self):
+        """
+        This role is the bare minimum required for a user to be able to have an account on
+        superset
+        """
+        hq_user_role = self.sm.add_role(HQ_USER_ROLE_NAME)
```
mkangia, line 162: nit: I wonder if this is causing unnecessary saves on this role in the db since it would exist once it has been created.
Non blocking but good to check and do a follow up if needed.

I was thinking if this should instead be added by AUTH_USER_ADDITIONAL_ROLES instead though I am not sure how that would work since its a custom role and not predefined in superset.

hq_superset/utils.py
```diff
@@ -120,16 +129,57 @@ class DomainSyncUtil:
...
+
+        hq_user_base_permissions = [
+            self.sm.add_permission_view_menu("can_profile", "Superset"),
+            self.sm.add_permission_view_menu("can_recent_activity", "Log"),
+        ]
+        self.sm.set_role_permissions(hq_user_role, hq_user_base_permissions)
+
+        if hq_user_role not in current_user.roles:
```
mkangia, line 170: nit: should this check be on the top so the role isn't set if its already present?

hq_superset/utils.py
```diff
@@ -139,29 +189,84 @@ def _ensure_schema_created(domain):
...
+            return {}, []
+
+        response_data = response.json()
+        hq_permissions = response_data['permissions']
+        roles = response_data['roles'] or []
+
+        # Map between HQ and CCA
+        permissions = {
+            CAN_WRITE_PERMISSION: hq_permissions["can_edit"],
+            CAN_READ_PERMISSION: hq_permissions["can_view"],
```
mkangia, line 237: nit: do we really need these mapping? we can just return permissions and that should do right?

hq_superset/views.py
```diff
@@ -40,6 +40,8 @@
 
 
 class HQDatasourceView(BaseSupersetView):
+    class_permission_name = "ExploreFormDataRestApi"
```
mkangia, line 43: just checking on why was this added?

hq_superset/views.py
```diff
@@ -63,6 +65,8 @@ def _ucr_id_from_pk(self, datasource_pk):
             return None
 
     @expose("/update/<datasource_id>", methods=["GET"])
+    @has_access
+    @permission_name("write")
```
mkangia, line 69: these checks were missing earlier?

mkangia (approved):
Hey @Charl1996 

Seems good to go ahead with this now.
 
Suggested some improvements, that are fine to taken as a follow up even as a separate ticket, if any of them seem reasonable. Considering the approach got simpler, which is great, code/implementation can be simplified further.

hq_superset/utils.py
```diff
@@ -120,16 +129,57 @@ class DomainSyncUtil:
...
-    def _ensure_schema_perm_created(self, domain):
-        menu_name = self.sm.get_schema_perm(get_hq_database(), get_schema_name_for_domain(domain))
-        permission = self.sm.find_permission_view_menu("schema_access", menu_name)
-        if not permission:
-            permission = self.sm.add_permission_view_menu("schema_access", menu_name)
-        return permission
+        The user gets assigned at least 3 roles in order to function on any domain:
+        1. hq_user_role: gives access to superset platform
+        2. domain_schema_role: restricts user access to specific domain schema
+        3. domain_user_role: restricts access for particular user on domain in accordance with how the permissions
```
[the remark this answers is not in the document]
Charl1996: Ah, good catch!

Charl1996:
@mkangia I'm going to merge this just to move it forward, but will be addressing your (very valid) comments in a followup PR.

hq_superset/views.py
Charl1996: Yes, but earlier we didn't need them. Now we need to set them to explicitly disable a read-only user from performing this action.

hq_superset/utils.py
Charl1996: I guess you're right. Will update (in upcoming PR)

hq_superset/utils.py
Charl1996: Mmm...I'll check it out

hq_superset/views.py
Charl1996: This was to ensure that only users with edit permissions can import data sources.

PR #60, opened 2024-09-03
Set default expiry for tokens

mkangia:


mkangia:
Closing in favor of https://github.com/dimagi/commcare-analytics/pull/62 which is the right way the expiry should be set

PR #61, opened 2024-09-05
Add superset command to subscribe to data source changes on HQ

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/SC-3910)

As part of the linked ticket it's necessary to trigger a (re)subscribe of the data sources on CCA with that on HQ. I've written a flask command to handle this task, so we can reuse this in the future if needed.

#### Approach
  I've decided to make the command reusable for various contexts, so we're now able to define the domain and data sources which needs subscribing in a normal text file on the server and let the command read from that to perform the necessary tasks. I think this approach would be most extensible in the long run, but I'm open to any other suggestions.

### DIFF
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -1,6 +1,6 @@
 import os
-
 import jinja2
+import click
 from flask import Blueprint
 
 
@@ -29,6 +29,8 @@ def flask_app_mutator(app):
     # it. So I've run out of better options. (Norman 2024-03-13)
     os.environ['AUTHLIB_INSECURE_TRANSPORT'] = '1'
 
+    register_commands(app)
+
 
 def override_jinja2_template_loader(app):
     # Allow loading templates from the templates directory in this project as well
@@ -49,3 +51,27 @@ def override_jinja2_template_loader(app):
     ))
     blueprint = Blueprint('Static', __name__, static_url_path='/static/images', static_folder=images_path)
     app.register_blueprint(blueprint)
+
+
+def register_commands(app):
+    @app.cli.command("subscribe-datasources")
+    @click.argument("superset_base_url")
+    @click.argument("hq_base_url")
+    @click.argument("data_sources_file_path")
+    @click.argument("username")
+    @click.argument("apikey")
+    def subscribe_all_data_sources(superset_base_url, hq_base_url, data_sources_file_path, username, apikey):
+        """
+        This command takes all data sources in the data_sources_file_path and subscribe to their changes on HQ.
+
+        Parameters:
+            superset_base_url: The base URL of the superset instance. This is used to construct the callback
+                URL for publishing the changes to.
+            hq_base_url: The base URL of the HQ server containing the data sources you'd like to subscribe to.
+            data_sources_file_path: Path to file containing JSON formatted data source IDs by domain,
+                e.g. {"domain": ["data_source_id"]}
+            username: An HQ superuser username; used for authorization.
+            apikey: An HQ superuser api key; used for authorization.
+        """
+        from hq_superset.commands import subscribe_data_sources
+        subscribe_data_sources(superset_base_url, hq_base_url, data_sources_file_path, username, apikey)
--- a/hq_superset/__init__.py
+++ b/hq_superset/__init__.py
@@ -60,7 +60,7 @@ def register_commands(app):
     @click.argument("data_sources_file_path")
     @click.argument("username")
     @click.argument("apikey")
-    def subscribe_all_data_sources(superset_base_url, hq_base_url, data_sources_file_path, username, apikey):
+    def subscribe_data_sources(superset_base_url, hq_base_url, data_sources_file_path, username, apikey):
         """
         This command takes all data sources in the data_sources_file_path and subscribe to their changes on HQ.
 
--- a/hq_superset/commands.py
+++ b/hq_superset/commands.py
@@ -24,7 +24,9 @@ def subscribe_data_sources(superset_base_url, hq_base_url, data_sources_file_pat
 
     failed_data_sources_by_id = {}
     for domain, datasource_ids in data_sources_by_domain().items():
+        print(f"Handling domain: {domain}")
         for datasource_id in datasource_ids:
+            print(f"Syncing datasource: {datasource_id}")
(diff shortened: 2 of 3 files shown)


hq_superset/commands.py
```diff
@@ -0,0 +1,55 @@
+import requests
+import json
+
+
+def subscribe_data_sources(superset_base_url, hq_base_url, data_sources_file_path, username, apikey):
```
mkangia, line 5: Can we make this possible without the file? say if we want to do for all? I assume its already subscribed, nothing will change?

mkangia: Resolved. We would like to be explicit instead.

mkangia, line 5: Hey @Charl1996 

Can you share more about whose username / apikey this is?
Are we using the same one for all domains?

I assumed we would use the domain's credentials like done in https://github.com/dimagi/commcare-analytics/blob/fb979e70b503e88cb101c66d8f52add281ce3a12/hq_superset/services.py#L173

Charl1996: @mkangia 
> Can you share more about whose username / apikey this is?

Any superuser. 

Since a dev is running this command I thought it's reasonable to expect the dev to be able to generate and use the superuser credentials. 

> I assumed we would use the domain's credentials like done in

Initially I used it by found out soon enough that there's problems with using the `HQRequest` class with respect to the auth token. The class relies on the fact that there's an active user request which is oauth authenticated. Since running a script manually won't have this context I decided to take this approach in this PR.

hq_superset/__init__.py
```diff
@@ -29,6 +29,8 @@ def flask_app_mutator(app):
     # it. So I've run out of better options. (Norman 2024-03-13)
     os.environ['AUTHLIB_INSECURE_TRANSPORT'] = '1'
 
+    register_commands(app)
```
kaapstorm, line 32: Instead of defining the command in two different places, you can use the same pattern that this `flask_app_mutator()` function uses to add views. For example ...

```suggestion
    from . import commands  # Append to import at the start of the function

    app.cli.add_command(commands.subscribe_datasources)
```

Then you can drop the `register_commands()` function below, and move the command definition, with decorators and docstring, to the `commands` module. e.g.
```python
# commands.py
import click


@click.command("subscribe-datasources")
@click.argument("superset_base_url")
@click.argument("hq_base_url")
@click.argument("data_sources_file_path", type=click.Path(exists=True))
@click.argument("username")
@click.argument("apikey", envvar="HQ_APIKEY")
def subscribe_datasources(
    superset_base_url,
    hq_base_url,
    data_sources_file_path,
    username,
    apikey,
):
    """
    This command takes all data sources in the data_sources_file_path
    and subscribes to their changes on HQ.

    :param superset_base_url: The base URL of the superset instance. This is
        used to construct the callback URL for publishing the changes to.

    :param hq_base_url: The base URL of the HQ server containing the data
        sources you'd like to subscribe to.

    :param data_sources_file_path: Path to file containing JSON-formatted
        data source IDs by domain. Example JSON:
        {"domain": ["data_source_id"]}

    :param username: An HQ superuser username.

    :param apikey: An HQ superuser API key. Alternatively, you can set the
        HQ_APIKEY environment variable to avoid passing this as an argument.
    """
    # params separated by blank lines for better readability on CLI:
    #     $ superset subscribe-datasources --help

```

hq_superset/commands.py
```diff
@@ -0,0 +1,55 @@
...
+            if not isinstance(ds_ids, list):
+                raise Exception(f"{domain} expected a list of data source IDs")
+        return datasource_ids_by_domain
+
+    webhook_url = f"{superset_base_url}/change/"
+    token_url = f"{superset_base_url}/token"
+
+    failed_data_sources_by_id = {}
+    for domain, datasource_ids in data_sources_by_domain().items():
+        print(f"Handling domain: {domain}")
```
kaapstorm, line 27: Not necessary, but recommended "[in case the terminal is misconfigured](https://click.palletsprojects.com/en/8.1.x/quickstart/#echoing)" :shrug: .
```suggestion
        click.echo(f"Handling domain: {domain}")
```

hq_superset/commands.py
```diff
@@ -0,0 +1,55 @@
...
+
+            endpoint = datasource_subscribe(domain, datasource_id)
+            data = {
+                'webhook_url': webhook_url,
+                'token_url': token_url,
+                'client_id': client.client_id,
+                'client_secret': client.get_client_secret(),
+            }
+            response = requests.post(
+                f"{hq_base_url}/{endpoint}",
```
kaapstorm, line 40: Use `urllib.parse.urljoin` to handle forward slashes gracefully. Django can figure out extra forward slashes, but proxies can make things unpredictable. Safer just to use `urljoin()`.

```suggestion
                urljoin(hq_base_url, endpoint),
```

hq_superset/commands.py
```diff
@@ -0,0 +1,55 @@
...
+                data=data,
+                headers={
+                    "Authorization": f"ApiKey {username}:{apikey}"
+                }
+            )
+
+            if response.status_code != 201:
```
kaapstorm, line 47: Nit: Just for the sake of defensive programming, in case some future dev changes HQ to subscribe asynchronously and return a 202 instead of a 201 -- who knows what future devs will think up? :upside_down_face: :joy: -- safer to check [all success status codes](https://docs.python-requests.org/en/latest/api/#requests.Response.ok):

```suggestion
            if not response.ok:
```

kaapstorm (changes requested):

hq_superset/commands.py
mkangia: We offlined about this and decided to skip re-subscription to avoid using a dev api key for this and additionally to have subscription authority with users/projects only.

kaapstorm: :+1: @Charl1996 does that mean that this PR will be closed?

Charl1996: @kaapstorm Thanks for flagging. Uhm...do we see any future use for something like this?

Charl1996:
See https://github.com/dimagi/commcare-analytics/pull/61#discussion_r1806420557

PR #65, opened 2024-10-08
Readme update related to pypi

ajeety4:
Minor readme updates regarding using version numbers for PyPi

### DIFF
--- a/apache-superset.md
+++ b/apache-superset.md
@@ -17,7 +17,7 @@ very complicated or impossible to do inside HQ Superset.
 
 In this spirit, 
 - if the change is specific to CommCare HQ, try to implement the
-  customization in [HQ Superset](https://github.com/dimagi/hq_superset).
+  customization in [Commcare Analytics](https://github.com/dimagi/hq_superset).
 - If the change is not specific to CommCare HQ and is useful to outside
   users, try to create a pull request against the upstream
   [Apache Superset](https://github.com/apache/superset) repo.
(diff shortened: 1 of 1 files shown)


apache-superset.md
```diff
@@ -306,7 +310,13 @@ the build instructions have remained the same before proceeding.
...
+  You can refer to 1Password for the PyPI API token for dimagi-superset.
 
 You should now be able to use this package inside
 https://github.com/dimagi/hq_superset by referring to the release tag.
+
+> Note: <ul><li>pypi package follows the versioning scheme as described [here](https://packaging.python.org/en/latest/discussions/versioning/).
+We make any initial customizations on top of apache upstream version and
+then make a release to pypi with the same version number. For e.g. customizations were made on the upstream version `3.1.0` and this was released to dimagi-superset pypi as `3.1.0`.<li>If any additional customizations are needed on top of initial versions, we are using the post release for numbering 
+i.e. `.post<>` as the suffix with N as an incrementing integer.
+It is acknowledged that using the post release is not the best way, however it was chosen to keep the version in sync to the upstream apache-superset version.</ul>
```
kaapstorm, line 322: You don't need to use `<ul><li>` tags in Markdown.

The following will be rendered the same way, and is more readable in plaintext.

```suggestion
> Note:
> 
> * PyPI package follows the versioning scheme as described
>   [here](https://packaging.python.org/en/latest/discussions/versioning/).
>   We make any initial customizations on top of Apache upstream version
>   and then make a release to PyPI with the same version number. e.g.
>   customizations were made on the upstream version `3.1.0` and this
>   was released to dimagi-superset PyPI as `3.1.0`.
>
> * If any additional customizations are needed on top of initial
>   versions, we are using the post release for numbering i.e.
>   `.post<>` as the suffix with N as an incrementing integer. It is
>   acknowledged that using the post release is not the best way,
>   however it was chosen to keep the version in sync with the upstream
>   apache-superset version.
```

kaapstorm (approved):

PR #69, opened 2024-10-15
Add talisman to example file

mkangia:
We added talisman config to superset setup.
This PR adds it to the example file.

Config shared with hints from how it is currently, as shared in https://github.com/dimagi/commcare-analytics-ansible/pull/31


Charl1996 (approved):
Good catch

PR #71, opened 2024-11-01
Assign gamma permissions to user domain role

Charl1996:
This PR is an alternative approach (see [this comment](https://github.com/dimagi/commcare-analytics/pull/59#discussion_r1825585233)) to how we assign "edit" permissions to a user on CCA. 

**Please comment on the preferred approach.**

### Glossary
`domain user role` -> the role assigned to a user on a domain which governs the user's permissions on the domain.

### What's changed?
The previous PR does the following:
1. For any view-only users we create a `domain user role` and give that role all the view-only permissions
2. For an edit user we don't create a new `domain user role` but rather give the standard `Gamma` role (we **don't** assign both the `domain user role` and the `Gamma` role since the `Gamma` role already has all the view permissions)
3. If a view-only user is upgraded to an edit user we need to make sure to clean up the redundant `domain user role` in the DB. 

This PR, however, does it this way:
1. For any user on the domain we create the `domain user role` regardless of what their permission level is. This role will be updated with the necessary permissions as is defined on HQ (view/edit)
2. If the user has only view access it works the same as the previous PR in that we only assign view permissions to the role
3. If the user has edit access we simply look at the `Gamma` role's permissions and assign those permissions to the role
4. If a view-only user is upgraded to an edit user we don't need any role clean-up, since the permissions is overwritten on the role itself.

The main difference between this PR and the previous one is that instead of assigning or unassigning the `Gamma` role / `domain user role` we assign the `Gamma` role permissions (or view-only permissions, depending on HQ configuration) always only to the **one role** (`domain user role`) which we use to manage the user's permissions.

### DIFF
--- a/hq_superset/tests/test_utils.py
+++ b/hq_superset/tests/test_utils.py
@@ -4,6 +4,7 @@ from unittest.mock import patch
 from hq_superset.utils import get_column_dtypes, DomainSyncUtil
 from .base_test import SupersetTestCase
 from .const import TEST_DATASOURCE
+from hq_superset.const import GAMMA_ROLE
 
 
 def test_get_column_dtypes():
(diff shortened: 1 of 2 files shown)


mkangia:
Thanks @Charl1996 

Just adding that in the first approach where we mention

> If a view-only user is upgraded to an edit user we need to make sure to clean up the redundant domain user role in the DB.

the opposite is also needed. In case of a downgrade, we need to remove gamma role assigned.

zandre-eng:
> For any view-only users we create a domain user role and give that role all the view-only permissions

I'm curious, is there a specific reason as to why we create a new role for every user in the domain? Would it not be possible to have a single "read-only" role per domain, and then assign all "read-only" users to the same role?

> Please comment on the preferred approach.

Whilst I do feel the approach from this PR does make the process of cleanup more straightforward when switching permissions for a user, I feel that this might make things more complex with regards to understanding which users have which permissions. Instead of simply looking at their role and being able to immediately see whether they are a "read-only" or "read-and-write" user, we would need to look at the permissions themselves to see what kind of user it is. This is also a good thing to note if we ever plan on having more granular permission access across a wider set of roles in CCA.

Given the above, I'm in favour of implementing role based access over permission based access, as it makes it easier to quickly identify which users have which set of permissions. In saying that however, we don't have a very varied set of permissions which users can have on CCA. It's either "read-only" or "read-and-write", so in that line of thinking the permission based approach should be okay if it's easier to implement and less code overhead.

Charl1996:
Actually, now that you've said this I'm more in favour of role-based access. 

If we stick to role-based access we can create a standard view-only role and simply assign that to users without creating a "domain user" read-only role for every user.

mkangia:
Great input about the single read only role @zandre-eng :100:

Charl1996:
This PR is not relevant anymore

PR #72, opened 2024-11-01
Expire user permissions

mkangia:
https://dimagi.atlassian.net/browse/SC-3475

Building up on work done in cs/SC-3473-user-roles-from-hq to expire user permissions in an hour to keep permissions up to date with HQ.

Review by commit :fish: 
There is quite some refactor, done in separate commits for clarity

All tests ran locally successfully. They wont run here since base branch isn't master.

https://dimagi.atlassian.net/browse/QA-7205

### DIFF
--- a/hq_superset/api.py
+++ b/hq_superset/api.py
@@ -11,9 +11,9 @@ from superset.views.base import (
     json_success,
 )
 
-from .models import DataSetChange
-from .oauth2_server import authorization, require_oauth
-from .tasks import process_dataset_change
+from hq_superset.models import DataSetChange
+from hq_superset.oauth2_server import authorization, require_oauth
+from hq_superset.tasks import process_dataset_change
 
 
 class OAuth(BaseApi):
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -3,6 +3,7 @@ from datetime import timedelta
 import superset
 from flask import current_app, flash, g, redirect, request, session, url_for
 from superset.config import USER_DOMAIN_ROLE_EXPIRY
+from superset.extensions import cache_manager
 
 from hq_superset.const import (
     SESSION_DOMAIN_ROLE_LAST_SYNCED_AT,
@@ -84,7 +85,9 @@ def sync_user_domain_role():
     ):
         return
     if _domain_role_expired():
-        return _sync_domain_role()
+        # only sync if another sync not in progress
+        if not _sync_in_progress():
+            return _perform_sync_domain_role()
 
 
 def _domain_role_expired():
@@ -94,6 +97,24 @@ def _domain_role_expired():
     time_since_last_sync = datetime_utcnow() - session[SESSION_DOMAIN_ROLE_LAST_SYNCED_AT]
     return time_since_last_sync >= timedelta(minutes=USER_DOMAIN_ROLE_EXPIRY)
 
+
+def _sync_in_progress():
+    return cache_manager.cache.get(_sync_domain_role_cache_key())
+
+
+def _sync_domain_role_cache_key():
+    return f"{g.user.id}_{g.hq_domain}_sync_domain_role"
+
+
+def _perform_sync_domain_role():
+    cache_key = _sync_domain_role_cache_key()
+
+    # set cache for 30 seconds
+    cache_manager.cache.set(cache_key, True, timeout=30)
+    sync_domain_role_response = _sync_domain_role()
+    cache_manager.cache.delete(cache_key)
+
+    return sync_domain_role_response
 
 def _sync_domain_role():
     if not DomainSyncUtil(superset.appbuilder.sm).sync_domain_role(g.hq_domain):
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -44,6 +44,7 @@ DOMAIN_EXCLUDED_VIEWS = [
     'AuthOAuthView.login',
     'AuthOAuthView.logout',
     'AuthOAuthView.oauth_authorized',
+    'CurrentUserRestApi.get_me',
     'DataSetChangeAPI.post_dataset_change',
     'OAuth.issue_access_token',
     'SelectDomainView.list',
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -1,16 +1,11 @@
-import superset
-
 from datetime import timedelta
-from flask import flash, g, redirect, request, session, url_for
 
+import superset
+from flask import flash, g, redirect, request, session, url_for
 from superset.config import USER_DOMAIN_ROLE_EXPIRY
 
-from .utils import (
-    SESSION_DOMAIN_ROLE_LAST_SYNCED_AT,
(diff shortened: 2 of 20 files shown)


hq_superset/tests/base_test.py
```diff
@@ -48,3 +49,33 @@ def tearDown(self):
                     sql = "; ".join(domain_schemas) + ";"
                     connection.execute(text(sql))
         super(HQDBTestCase, self).tearDown()
+
+
+class LoginUserTestMixin(object):
```
Charl1996, line 55: Nice. Seems useful.

hq_superset/utils.py
```diff
@@ -415,3 +418,7 @@ def cast_data_for_table(
 def generate_secret():
     alphabet = string.ascii_letters + string.digits
     return ''.join(secrets.choice(alphabet) for __ in range(64))
+
+
+def datetime_utcnow_naive():
```
Charl1996, line 420: @mkangia 
I think this function name is a misnomer. You're returning an [aware datetime](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects), are you not?.

hq_superset/hq_domain.py
```diff
@@ -64,6 +77,28 @@ def ensure_domain_selected():
...
+        return
+    if _domain_role_expired():
+        _sync_domain_role()
+
+
+def _domain_role_expired():
+    if not session.get(SESSION_DOMAIN_ROLE_LAST_SYNCED_AT):
+        return True
+
+    time_since_last_sync = datetime_utcnow_naive() - session[SESSION_DOMAIN_ROLE_LAST_SYNCED_AT]
```
Charl1996, line 93: @mkangia 
Nit: for more robust implementation we could check if `session[SESSION_DOMAIN_ROLE_LAST_SYNCED_AT]` is a datetime-like object that could be operated on. 

How about:
```
last_sync_datetime = session[SESSION_DOMAIN_ROLE_LAST_SYNCED_AT]
if not isinstance(last_sync_datetime, datetime):
    return False
```

hq_superset/hq_domain.py
```diff
@@ -64,6 +77,28 @@ def ensure_domain_selected():
...
+    return time_since_last_sync >= timedelta(minutes=USER_DOMAIN_ROLE_EXPIRY)
+
+
+def _sync_domain_role():
+    DomainSyncUtil(superset.appbuilder.sm).sync_domain_role(g.hq_domain)
```
Charl1996, line 98: @mkangia 
What happens if the domain role cannot sync during this routine? Is it fine to keep the user logged in in the current state?

hq_superset/hq_domain.py
```diff
@@ -44,6 +44,7 @@ def after_request_hook(response):
     'AuthOAuthView.login',
     'AuthOAuthView.logout',
     'AuthOAuthView.oauth_authorized',
+    'CurrentUserRestApi.get_me',
```
Charl1996, line 47: Curious what this is for?

Charl1996:
@mkangia Nice! Thanks for the refactors and good tests!

hq_superset/utils.py
mkangia: You are right, I got that backwards, will rename this `datetime_utcnow`.

hq_superset/hq_domain.py
mkangia: Cool. I can add that. This value is only always set by CCA and can't edited by the user so it would certainly be a datetime but no harm in adding safety.

hq_superset/tests/test_views.py
```diff
@@ -134,7 +134,7 @@ def _do_assert(datasources):
         client.get('/hq_datasource/list/', follow_redirects=True)
         _do_assert(self.oauth_mock.test2_datasources)
 
-    @patch.object(DomainSyncUtil, "sync_domain_role", return_value=True)
+    @patch.object(DomainSyncUtil, "_get_domain_access", return_value=({"can_write": True, "can_read": True}, []))
```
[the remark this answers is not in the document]
mkangia: It was a flaky test that succeeded when run separately or if only this test file was run but failed when the full test suite was run.
I believe the permissions were getting set unexpectedly or getting revoked somehow, so I set them specifically instead of skipping it via the patch for `sync_domain_role`. I did try to debug it for a couple of hours before letting it be.

hq_superset/hq_domain.py
mkangia: From the UI, this a call made to constantly sync the user record, I assume its getting used in javascript/UI somewhere.
It is a domain independent API call.
I added it to this list so we don't fire the before request hooks on such requests since they are not needed.

hq_superset/hq_domain.py
mkangia: This turned out to be a little challenging, but I got it working like this
https://github.com/dimagi/commcare-analytics/pull/72/commits/a1959294970440ed801501d651b7e9a64da8d255

Screenshot from 2024-11-05 00-58-37

hq_superset/hq_domain.py
mkangia: Found one more, https://github.com/dimagi/commcare-analytics/pull/72/commits/b4eb1b3b1745278ea700f8fa5c98bf9109fa09b2

hq_superset/api.py
```diff
@@ -11,9 +11,9 @@
     json_success,
 )
 
-from .models import DataSetChange
-from .oauth2_server import authorization, require_oauth
-from .tasks import process_dataset_change
+from hq_superset.models import DataSetChange
```
ajeety4, line 14: :+1:

ajeety4 (commented):
Great Refactors.  The changes looks good. 

Just a query related to the fix made in the [commit](https://github.com/dimagi/commcare-analytics/pull/72/commits/716651c843daba4cf17da0aa0507b16ce6746365)
> - /api/v1/me is called frequently leading to before_request_hook getting fired unnecessarily, specifically for sync_user_domain_role leading to more than one syncs happening at once leading to db conflicts/errors

Does the db conflicts/errors happened because of API calls that are simultaneous?  Wondering if the same thing can occur if a user is working with multiple tabs at once ?

Charl1996 (approved):
Thanks for the changes! Looks good to me!

mkangia:
Hey @ajeety4 

> Does the db conflicts/errors happened because of API calls that are simultaneous? Wondering if the same thing can occur if a user is working with multiple tabs at once ?

Spot on. That fix is where I noticed it first.
I made the [following change](https://github.com/dimagi/commcare-analytics/pull/72/commits/a5f2c58adc7f1ba107b6777ea0b77d0f07ce3e52) in order to limit to only one sync at once.
It does appear to have solved the issue in most cases. 

However there can be some weird edge cases where two requests come exactly at the same moment which I am letting go for now since sync will get reattempted at the next request if needed for such edge cases.
I noticed that locally but not on staging during some minor testing.

mkangia:
Thanks for the reviews @Charl1996

ajeety4 (approved):

ajeety4 (approved):

PR #73, opened 2024-11-03
Update staging.yml

mkangia:
Since master branch is protected we really can't push code to it as described in https://github.com/dimagi/commcare-analytics?tab=readme-ov-file#testing-on-staging


Charl1996 (approved):

Charl1996:
@mkangia Did you update the repo to mark the `master` branch as protected?

mkangia:
Hey @Charl1996 
I am not sure I did that but it is something we should keep.

PR #79, opened 2024-11-15
Post PR updates

Charl1996:
This PR addresses the promised changes [here](https://github.com/dimagi/commcare-analytics/pull/59#issuecomment-2459537287) and does a small refactor to simplify the code.


hq_superset/tests/test_utils.py
```diff
@@ -122,7 +122,4 @@ def _ensure_platform_roles_exist(self, sm):
 
     @staticmethod
     def _to_permissions_response(can_write, can_read, roles):
-        return {
-            "can_write": can_write,
-            "can_read": can_read,
-        }, roles
+        return can_read, can_write, roles
```
Charl1996, line 125: This method seems redundant, but it's providing a more clear structure to the expected results in the test. 

For instance, instead of doing this in the test
```
  get_domain_access_mock.return_value = (False, False ,["sql_lab", "dataset_editor"])
```
I can do this
```
get_domain_access_mock.return_value = self._to_permissions_response(
    can_write=False,
    can_read=False,
    roles=["sql_lab", "dataset_editor"],
)
```
Using a method like this makes it more clear what each value means.

hq_superset/utils.py
```diff
@@ -135,10 +135,7 @@ def sync_domain_role(self, domain):
         The user gets assigned at least 3 roles in order to function on any domain:
         1. hq_user_role: gives access to superset platform
         2. domain_schema_role: restricts user access to specific domain schema
-        3. domain_user_role: restricts access for particular user on domain in accordance with how the permissions
-        are defined on CommCare HQ.
-
-        Any additional roles defined on CommCare HQ will also be assigned to the user.
+        3. Either the Gamma role for "edit" users or the READ_ONLY_ROLE_NAME for "view only" user
```
mkangia, line 136: nit: what is an "edit" user? or "view only" user? in terms of CCA

hq_superset/utils.py
```diff
@@ -199,19 +197,19 @@ def _ensure_domain_role_created(self, domain):
...
 
-        if domain_permissions[CAN_WRITE_PERMISSION]:
+        if can_write:
             user_role = GAMMA_ROLE_NAME
         else:
             self._ensure_read_only_role_exists()
             user_role = READ_ONLY_ROLE_NAME
 
-        roles_names.append(user_role)
+        platform_roles_names.append(user_role)
```
mkangia, line 210: I'd suggest to avoid editing an mutable object. This can cause confusion since the original object is being edited here.
Rather make a new list using platform_roles_names and then append to it.

mkangia (commented):
Very well.

Just one blocking feedback which is easy to fix, if needed.

hq_superset/utils.py
```diff
@@ -228,17 +226,7 @@ def _get_domain_access(domain):
...
-            CAN_WRITE_PERMISSION: hq_permissions["can_edit"],
-            CAN_READ_PERMISSION: hq_permissions["can_view"],
-        }
-        return permissions, roles
-
-    @staticmethod
-    def _user_has_no_access(permissions: dict):
-        user_has_access = any([permissions[p] for p in permissions])
-        return not user_has_access
+        return hq_permissions["can_view"], hq_permissions["can_edit"], roles
```
kaapstorm, line 229: Previous `return` on line 223 returns a dict and a list, but this returns two bools (?) and a list. It looks like you need to update line 223.

kaapstorm: A test that returns a status code other than 200 would be cool, because it would have flagged the problem.

hq_superset/utils.py
kaapstorm: :heavy_plus_sign: :one:

hq_superset/utils.py
mkangia: oh, great catch @kaapstorm

Charl1996: @kaapstorm 

Will update!

Charl1996: https://github.com/dimagi/commcare-analytics/pull/79/commits/ca7285d967dfb9790ad639442a70ad58b835a7be

hq_superset/utils.py
Charl1996: https://github.com/dimagi/commcare-analytics/pull/79/commits/52366715d1204ce01747e524ee8149eb82539958

hq_superset/utils.py
Charl1996: https://github.com/dimagi/commcare-analytics/pull/79/commits/13d513caad58b27266b34a072e262e00b4c4a700

kaapstorm (approved):

hq_superset/utils.py
kaapstorm: Nice! Thank you.

PR #81, opened 2024-11-21
Update README with staging deploy steps

Charl1996:
This PR is updating the readme with the staging deploy steps as outlined [here](https://dimagi.atlassian.net/wiki/spaces/GS/pages/2144316885/CommCare+Analytics+Deployment).

The reason for this is mainly so that whoever is reading the "Testing on staging" part in the readme don't have to search for an additional doc on how to deploy staging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ In order to test your feature branch on staging you need to
 6. Deploy `staging` (see Deploying Staging below)
 
 #### Deploying staging
-1. SSH to `commcare-analytics-staging.dimag.com`
+1. SSH to `commcare-analytics-staging.dimagi.com`
 2. Virtual environment should already be activated
 3. Uninstall current superset version
 
--- a/README.md
+++ b/README.md
@@ -156,10 +156,33 @@ don't include Superset.
 In order to test your feature branch on staging you need to
 1. Check out to `master` branch (make sure it's up to date)
 2. Add your feature branch to the `branches` section in `scripts/staging.yml` file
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -156,10 +156,33 @@ don't include Superset.
...
-3. Push the file back to remote `master`
-4. Run `scripts/rebuildstaging` (this will rebuild the staging branch on your machine)
+3. Make a PR to `master`
+4. Once merged, run `scripts/rebuildstaging` (this will rebuild the staging branch on your machine)
 5. Push the new `staging` branch to remote
-6. Deploy `staging` as usual
+6. Deploy `staging` (see Deploying Staging below)
+
+#### Deploying staging
+1. SSH to `commcare-analytics-staging.dimag.com`
```
zandre-eng, line 165:
```suggestion
1. SSH to `commcare-analytics-staging.dimagi.com`
```

Charl1996: Nice catch

zandre-eng (approved):

mkangia (approved):
I believe we eventually need something like https://github.com/dimagi/staging-branches/ for CCA or can we use that same repo itself for cca?

Charl1996:
> I believe we eventually need something like https://github.com/dimagi/staging-branches/ for CCA or can we use that same repo itself for cca?

Yes, I think so. 

I've added [this ticket](https://dimagi.atlassian.net/browse/SC-4084) yesterday, so maybe it can form part of that.

PR #90, opened 2024-12-15
Add task to clean up shared files regularly

mkangia:
https://dimagi.atlassian.net/browse/SC-4051

### DIFF
--- a/hq_superset/tasks.py
+++ b/hq_superset/tasks.py
@@ -1,4 +1,6 @@
 import os
+import superset
+import time
 
 from superset.extensions import celery_app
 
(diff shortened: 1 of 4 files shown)


Charl1996 (approved):
Looks good! 

Curious why you didn't opt for a server cron task? 

Just to be clear, I don't have a problem with the task in the code (in fact, it's easier to maintain I think), but just wondering how you thought about it.

mkangia:
Hey @Charl1996 

Ya, good question.

It was definitely the maintainability as you mentioned.
Additionally visibility that the files are removed by having the task in this code instead of the ansible repo, plus, having it easily configurable here via superset config.

But overall,
I would think of log files or such to be cleaned via cron but files generated by the application to be removed by application itself.

Charl1996:
>  files generated by the application to be removed by application itself

That's a good point

mkangia:
A bug in setting up the task, resolved in https://github.com/dimagi/commcare-analytics/pull/91

PR #95, opened 2024-12-19
Configure skipping data change for domains

mkangia:
A way to configure any domains to be skipped during data source processing.

This would come in handy and I currently need it to skip changes queued for a project I would manually refresh making the changes redundant.


hq_superset/models.py
```diff
@@ -41,6 +51,10 @@ def update_dataset(self):
             raise TableMissing(f'{self.data_source_id} table not found.')
         table = sqla_table.get_sqla_table_object()
 
+        if table.schema and table.schema in SKIP_DATASET_CHANGE_FOR_SCHEMAS:
```
zandre-eng, line 54: Will this then only ignore future dataset changes, or will it also ignore ones that are currently queued?

zandre-eng (approved):

hq_superset/models.py
zandre-eng: Had an offline sync with Manish.

mkangia: Offlined that the change will apply to any domains added to config and celery restarted.

PR #98, opened 2025-01-13
Update staging.yml

kaapstorm:
For staging new maps.

Jira: [SC-3941](https://dimagi.atlassian.net/browse/SC-3941)


kaapstorm:
There is only one branch to be staged. So I just deployed that branch. Closing because this change isn't needed.

PR #99, opened 2025-01-13
Test maps

kaapstorm:
Draft PR just to show the changes to be staged.

Do not merge.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import find_packages, setup
 
 setup(
     name='commcare-analytics',
-    version='1.1.2',
+    version='1.1.2.dev0',
     description='CommCare HQ Superset Integration',
     license='Apache2',
     author='Dimagi Inc.',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,10 +11,10 @@
     packages=find_packages(exclude=['docs', 'tests']),
     include_package_data=True,
     install_requires=[
-        'dimagi-superset==3.1.0.post1',
+        'dimagi-superset @ git+https://github.com/dimagi/superset.git@e0aa79525a',
         # Dependencies based on Superset 3.1.0 where applicable
         'Authlib==1.3.0',
-        'celery==5.2.7',
+        'celery==5.4.0',
```
mkangia, line 17: just checking that this change was intentional? we would be upgrading celery?

kaapstorm: This is just so that the Celery requirement of CommCare Analytics doesn't cause an unresolvable conflict with the [Celery requirement](https://github.com/dimagi/superset/blob/master/setup.py#L77) of Dimagi Superset (`"celery>=5.3.6, <6.0.0"`).

mkangia: is that the version currently installed on the environments (staging/production)?

kaapstorm: It's currently installed on Staging. I have not checked Prod.

PR #100, opened 2025-01-14
Datadog application monitoring

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/SC-4106)

This PR 
1. adds `datadog` to the codebase
2. Sends a timed metric to datadog

The way it works is, when you're invoking `statd` from `datadog` it send the metric to the `datadog agent` running on the machine (assuming it's running), which in turn handles the authentication etc. Since the datadog agent is already present on the `staging` and `production` machines this should just work, but I'll update here what my findings are.
 
I've considered simply adapting HQ's implementation of the metrics, but decided against that because we really don't need all that code at this stage and can get away with a clearly much simpler approach.

The corresponding CCA Ansible change can be found [here](https://github.com/dimagi/commcare-analytics-ansible/pull/48).


kaapstorm (approved):

superset_config.example.py
```diff
@@ -191,3 +191,5 @@ class CeleryConfig:
 
 USER_DOMAIN_ROLE_EXPIRY = 60 # minutes
 SKIP_DATASET_CHANGE_FOR_DOMAINS = []
+
+SERVER_ENVIRONMENT = 'changeme'  # staging, production, etc.
```
ajeety4, line 195: Assuming this would be set up by ansible ?

ajeety4 (approved):
LGTM

hq_superset/models.py
```diff
@@ -34,6 +35,10 @@ class DataSetChange:
     data: list[dict[str, Any]]
 
     def update_dataset(self):
+        with statsd.timed('cca.dataset_change.timer', tags=[self.data_source_id]):
```
mkangia: :fire:

mkangia: and :moneybag: ?

superset_config.example.py
mkangia: :+1:

hq_superset/models.py
```diff
@@ -35,7 +35,8 @@ class DataSetChange:
     data: list[dict[str, Any]]
 
     def update_dataset(self):
-        with statsd.timed('cca.dataset_change.timer', tags=[SERVER_ENVIRONMENT, self.data_source_id]):
+        env_tag = f"env:{SERVER_ENVIRONMENT}"
+        with statsd.timed('cca.dataset_change.timer', tags=[env_tag, f"datasource:{self.data_source_id}"]):
```
mkangia, line 39: can we add domain tag ? I guess that would be useful?

Will datasource show on the metric itself and env would be on the top in the dropdown? and then if we add domain it would also be on the top of the dashboard?

mkangia:
Just love it how straight forward this was and +1 on keeping things simple and to what we need.

mkangia:
Will wait for your findings for dropping an approval.

hq_superset/models.py
Charl1996: I'm currently busy figuring that out, since this would potentially result in massive volumes.

hq_superset/models.py
mkangia: I don't know how tags are used/available on datadog once they are added here in metrics.
For now, I have seen "env" or "domain" being available on top of a datadog dashboard (like [here](https://app.datadoghq.com/dashboard/65q-24f-36f/asynchronous-ucr-dashboard?fromUser=false&refresh_mode=sliding&tpl_var_env%5B0%5D=production&from_ts=1737008909727&to_ts=1737009809727&live=true)) and then additionally there are metrics available for different items within a graph (like [here](https://app.datadoghq.com/dashboard/65q-24f-36f/asynchronous-ucr-dashboard?fromUser=false&fullscreen_end_ts=1737009866111&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_section=overview&fullscreen_start_ts=1737008966111&fullscreen_widget=219512687&refresh_mode=sliding&tpl_var_env%5B0%5D=production&from_ts=1737008909727&to_ts=1737009809727&live=true))
So, I was asking if the domain and data source ids will be available like so.

Charl1996: Ah, I see. I created a chart [here](https://us5.datadoghq.com/dashboard/9x3-a9n-zts/datasource-metrics?fromUser=false&fullscreen_end_ts=1737017307919&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_section=edit&fullscreen_start_ts=1737013707919&fullscreen_widget=2249605656428426&refresh_mode=sliding&from_ts=1737013855744&to_ts=1737017455744&live=true) to demonstrate that we can, yes.

(I'm not 100% sure if the timeseries is what we want long term, maybe the Top List chart works better?)

mkangia: Love it, thank you!

mkangia: timeseries seems okay from the little I know, it does give me the data I think I want.

Charl1996:
I just want to touch up one or two things before I'll open this for review later today.

Charl1996:
Datadog has been tested on staging and the metrics pulls through (see [cca.dataset_change.timer](https://us5.datadoghq.com/metric/explorer?fromUser=false&start=1737361289731&end=1737364889731&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGABuMMiZmRgAdMkIGFoITpmMGCLw9Wo69QCcAKyKWXAiEmhwwABUPCA8ALpUru54mKHhq6rrGDGl8QvLIBpyUzIg8hhTCAg5STgw7RsaEJmJaKJwTnKKyumfKAfL5OehMZQiNweNALfgQeyYLA-BT3T4iJRLHh8E7yT4IADCUmEMBQInWaB4QA) in metrics explorer).

Regarding costs:
There's not a high cardinality with this particular metric (currently 33 DataSource repeaters on production, including test ones). This puts us well below the "$5-per-100-custom-metrics" margin.

I've also spoken to Graham and his sentiment is that we don't have to worry too much about the volume of this metric (ie. the amount of potential data points per custom metric).

superset_config.example.py
zandre-eng: @Charl1996 Just checking whether we will still be using this given that we do have the hostname set up in the Datadog config with the Ansible setup?

zandre-eng (approved):
Looks good.

superset_config.example.py
Charl1996: The `superset_config.example.py` file is still used when we set up for local development. The code will expect it to be present since it's importing it.

ajeety4 (approved):

PR #104, opened 2025-01-14
Fix order in staging.yml

kaapstorm:
No rush on this one, because I lost patience and rebuilt the staging branch manually. :expressionless: 

But I was an idiot, and put the merge branch _after_ the conflicting branches, instead of _before_ them. This PR has the correct order.

I'm opening the PR so that `staging.yml` in master is an accurate reflection of what is deployed.

### DIFF
--- a/scripts/staging.yml
+++ b/scripts/staging.yml
@@ -7,6 +7,6 @@
 trunk: master
 name: staging
 branches:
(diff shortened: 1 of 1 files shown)


Charl1996 (approved):

Charl1996:
> put the merge branch after the conflicting branches, instead of before them

I didn't catch that subtlety either...comes from rushing a reviewing

PR #107, opened 2025-01-22
Bump version

Charl1996:
Bump version in preparation for deploy 23 Jan @ 18h00 UTC. 

I will merge this just before the deploy.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import find_packages, setup
 
 setup(
     name='commcare-analytics',
-    version='1.1.2',
+    version='1.2.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='commcare-analytics',
-    version='1.1.2',
+    version='1.2.0',
```
mkangia, line 5: I assume we did 1.2.0 and not 1.1.3 because this is not a bug fix and a feature addition which warrants the minor version to be updated?

Charl1996: Correct.

mkangia (approved):

PR #118, opened 2025-02-25
Error Handling for False-Positive SQLLab Exception

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/SC-4284).

This PR introduces a small change in which an extra check is added to the `after_request` handler for the Superset Flask application. This new hook essentially checks for 500 requests coming in from the `/api/v1/sqllab/execute/` endpoint which have a certain error type and changes them to 400.

The reason for this change is that SQLLab allows for creating invalid queries (e.g. accessing a table that doesn't exist). When this happens, Superset simply throws an unhandled 500 exception which then gets raised as a false-positive 500 request on Datadog. Changing the status code to 400 more appropriately flags it as simply an invalid request rather than an exception.

From a user's perspective nothing is changing and they will still see the same error message on the front-end:
Screenshot from 2025-03-13 15-42-31

But the logs will now reflect the error as a 400 instead of a 500:

~~Please note, that there exists a related PR in the commcare-analytics-ansible repo to add the same to the superset_config file for an Ansible deploy. The PR can be accessed [here](https://github.com/dimagi/commcare-analytics-ansible/pull/54).~~

### DIFF
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -63,9 +63,12 @@ def error_check_after_request(response_obj):
             'GENERIC_DB_ENGINE_ERROR',
         ]
         response_dict = json.loads(response.decode('utf-8'))
-        for error in response_dict['errors']:
-            if error['error_type'] in VALID_ERROR_TYPES:
-                return True
+        try:
+            for error in response_dict['errors']:
+                if error['error_type'] in VALID_ERROR_TYPES:
+                    return True
+        except KeyError:
+            pass
         return False
 
     if (
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -57,7 +57,7 @@ def error_check_after_request(response_obj):
     endpoints that return false-positive 500 errors and change them to
     a more appropriate 400 error.
     """
-    def has_valid_errors(response):
+    def should_update_status_code_for_errors(response):
         VALID_ERROR_TYPES = [
             'COLUMN_DOES_NOT_EXIST_ERROR',
             'GENERIC_DB_ENGINE_ERROR',
@@ -71,7 +71,7 @@ def error_check_after_request(response_obj):
     if (
         request.path.startswith("/api/v1/sqllab/execute/")
         and response_obj.status_code == 500
-        and has_valid_errors(response_obj.response[0])
+        and should_update_status_code_for_errors(response_obj.response[0])
     ):
         response_obj.status_code = 400
     return response_obj
--- a/hq_superset/hq_domain.py
+++ b/hq_superset/hq_domain.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+import json
 
 import superset
 from flask import current_app, flash, g, redirect, request, session, url_for
@@ -37,6 +38,7 @@ def after_request_hook(response):
     ]
     if request.url_rule and (request.url_rule.endpoint in logout_views):
         response.set_cookie('hq_domain', '', expires=0)
+    response = error_check_after_request(response)
     return response
 
 
@@ -47,6 +49,32 @@ def oauth_session_expired(*arg, **kwargs):
         'warning',
     )
     return redirect(url_for("AuthOAuthView.login"))
+
+
+def error_check_after_request(response_obj):
+    """
+    This gets used by the Flask app's after_request hook to check for
+    endpoints that return false-positive 500 errors and change them to
+    a more appropriate 400 error.
+    """
+    def has_valid_errors(response):
+        VALID_ERROR_TYPES = [
+            'COLUMN_DOES_NOT_EXIST_ERROR',
+            'GENERIC_DB_ENGINE_ERROR',
+        ]
+        response_dict = json.loads(response.decode('utf-8'))
+        for error in response_dict['errors']:
+            if error['error_type'] in VALID_ERROR_TYPES:
+                return True
+        return False
+
+    if (
+        request.path.startswith("/api/v1/sqllab/execute/")
+        and response_obj.status_code == 500
+        and has_valid_errors(response_obj.response[0])
+    ):
+        response_obj.status_code = 400
+    return response_obj
 
 
 DOMAIN_EXCLUDED_VIEWS = [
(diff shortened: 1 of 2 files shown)


superset_config.example.py
```diff
@@ -192,4 +193,20 @@ class CeleryConfig:
...
 SKIP_DATASET_CHANGE_FOR_DOMAINS = []
 
-SERVER_ENVIRONMENT = 'changeme'  # staging, production, etc.
\ No newline at end of file
+SERVER_ENVIRONMENT = 'changeme'  # staging, production, etc.
+
+def FLASK_APP_MUTATOR(app: Flask):
+    app.after_request(error_check_after_req)
+
+def error_check_after_req(response_obj):
```
zandre-eng, line 201: I tried moving this function to a different file to keep things a bit cleaner, however this causes a no application context error to be thrown when this function is called. Moving it back to the config where `FLASK_APP_MUTATOR` is called fixes the problem.

superset_config.example.py
```diff
@@ -192,4 +193,20 @@ class CeleryConfig:
...
+    """
+    This gets used by the Flask app's after_request hook to check for
+    endpoints that return false-positive 500 errors and change them to
+    a more appropriate 400 error.
+    """
+    if (
+        request.path.startswith("/api/v1/sqllab/execute/")
+        and response_obj.status == "500 INTERNAL SERVER ERROR"
```
Charl1996, line 209: Can't we get access to only the status code, ie. `500`? I get uneasy when we have to compare strings.

Charl1996:
Nice! Just a small comment.

superset_config.example.py
zandre-eng: Okay I did a bit more digging into the classes which the `Response` class inherits from and it seems there is a `status_code` property we can use! 

Have updated in [a469f33](https://github.com/dimagi/commcare-analytics/pull/118/commits/a469f33f3c0440c16005c9f587254df9577d734c). I agree strings are always a messy business.

ajeety4 (approved):

superset_config.example.py
mkangia: Same question here as in the related PR, https://github.com/dimagi/commcare-analytics-ansible/pull/54/files#r1970329777

zandre-eng: Please see response in related PR [here](https://github.com/dimagi/commcare-analytics-ansible/pull/54/files#r1971106425).

zandre-eng:
I have done a small refactor with [7dbcb1c](https://github.com/dimagi/commcare-analytics/pull/118/commits/7dbcb1c0a9bccc8a69a0a8f6a8f28b580fba2ad3) to move the check function from the config into `hq_domain.py`. 

I noticed that putting it in the config was actually overriding the `after_request` hooks we had in `hq_domain.py`. Moving it allows us to do the check while still retaining the other after_request` hooks.

Additionally, I also refactored the check slightly to look for specific error types in the request. We will only change the response status code if it matches any of the error types from the list.

hq_superset/hq_domain.py
```diff
@@ -49,6 +51,32 @@ def oauth_session_expired(*arg, **kwargs):
     return redirect(url_for("AuthOAuthView.login"))
 
 
+def error_check_after_request(response_obj):
+    """
+    This gets used by the Flask app's after_request hook to check for
+    endpoints that return false-positive 500 errors and change them to
+    a more appropriate 400 error.
+    """
+    def has_valid_errors(response):
```
Charl1996, line 60: @zandre-eng Can we rename this function to `can_ignore_errors` or `can_deescalate_error_response` because signalling that the error is "valid" feels like it is indeed an error and should give error response. Or maybe I'm thinking too hard about this? (feel free to differ)

zandre-eng: I think that's a valid point. How does something like `has_db_query_error` sound? Since we're essentially checking whether the user's SQL query raised a specific error type.

mkangia: agreed that this could use a better name, I can suggest
`update_status_code_for_errors`

mkangia (approved):
Thanks for the follow up @zandre-eng

mkangia:
Can you share in the PR description what users will see now?

hq_superset/hq_domain.py
zandre-eng: I like that suggestion. I updated it to `should_update_status_code_for_errors` to better reflect that this is returning a boolean ([03b8dd3](https://github.com/dimagi/commcare-analytics/pull/118/commits/03b8dd3d4d4e9e8f7bd0950e4f64ee045a5cb9ea)).

zandre-eng:
> Can you share in the PR description what users will see now?

Hey @mkangia nothing will change from a user's perspective, and they will still see the same error message on the front-end. But I have updated the description with screenshots to reflect this, as well as what the logs look like now.

mkangia:
Thank you @zandre-eng, that works.

mkangia (approved):

hq_superset/hq_domain.py
```diff
@@ -49,6 +51,32 @@ def oauth_session_expired(*arg, **kwargs):
...
+        VALID_ERROR_TYPES = [
+            'COLUMN_DOES_NOT_EXIST_ERROR',
+            'GENERIC_DB_ENGINE_ERROR',
+        ]
+        response_dict = json.loads(response.decode('utf-8'))
+        for error in response_dict['errors']:
```
mkangia, line 66: Is it possible that there won't be "errors" key in which case this would error out?

mkangia: Oh I see that this is being called in case of a 500 only.
I leave it to you to decide if we should just check for "errors" key to be present first.

mkangia (approved):

hq_superset/hq_domain.py
zandre-eng: From my local testing the "errors" key is always present when there is a 500, but I appreciate the cautious thinking. Rather be careful than sorry, especially since things can change in the future or there might be an unexpected edge case. Will add a check in here.

mkangia (approved):

